PR #439, opened 2023-05-01
Fix product owner mentions

chadwhitacre:
- removes label description hack (since we have better control of both sides now)
- standardize on `product-owners-` prefix
- match slugification algorithm to [`security-as-code`](https://github.com/getsentry/security-as-code/blob/fb50e07387ad8ab7724a5dfda4eabefcce6a28da/rbac/lib/make-product-owners#L13-L16)

### DIFF
--- a/src/api/github/__mocks__/getClient.ts
+++ b/src/api/github/__mocks__/getClient.ts
@@ -56,7 +56,10 @@ function mockClient() {
     },
     teams: {
       getByName: jest.fn(async function (payload) {
-        if (payload.team_slug === 'test' || payload.team_slug === 'rerouted') {
+        if (
+          payload.team_slug === 'product-owners-test' ||
+          payload.team_slug === 'product-owners-rerouted'
+        ) {
           return { status: 200, data: {} };
         }
         throw new MockOctokitError(404);
--- a/src/utils/slugizeProductArea.test.ts
+++ b/src/utils/slugizeProductArea.test.ts
@@ -1,0 +1,21 @@
+import { slugizeProductArea } from './slugizeProductArea';
+
+describe('slugizeProductArea', function () {
+  describe('slugize tests', function () {
+    it('behaves as expected', async function () {
+      const cases = [
+        ['MOST of ALL', 'most-of-all'],
+        ['Cheese & Bread', 'cheese-bread'],
+        ['otherwise - notherwise', 'otherwise-notherwise'],
+        ['over - & - yonder', 'over-yonder'],
+        ["other's druthers", 'others-druthers'],
+      ];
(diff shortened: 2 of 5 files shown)


src/api/github/__mocks__/getClient.ts
```diff
@@ -56,7 +56,10 @@ function mockClient() {
     },
     teams: {
       getByName: jest.fn(async function (payload) {
-        if (payload.team_slug === 'test' || payload.team_slug === 'rerouted') {
+        if (
+          payload.team_slug === 'product-owners-test' ||
+          payload.team_slug === 'product-owners-rerouted'
```
hubertdeng123, line 61: is it more clear to change this to `product_owner_team_slug`? since there are different types of teams now

src/utils/slugizeProductArea.ts
```diff
@@ -0,0 +1,15 @@
+export function slugizeProductArea(s) {
+  // Keep this in sync with slugize in ...
+  // https://github.com/getsentry/security-as-code/blob/main/rbac/lib/make-product-owners
```
hubertdeng123, line 3: wonder if at a certain point it's possible to use changes in `security-as-code` to add/remove labels since there's a lot of hoops to jump through to add a new product area

hubertdeng123 (approved):

src/utils/slugizeProductArea.ts
chadwhitacre: That would be great! :)

src/api/github/__mocks__/getClient.ts
chadwhitacre: This is driven by GitHub's API design. Where we call this in `route.ts` I prefixed with `gh` to clarify.

src/utils/slugizeProductArea.test.ts
```diff
@@ -0,0 +1,21 @@
+import { slugizeProductArea } from './slugizeProductArea';
+
+describe('slugizeProductArea', function () {
+  describe('slugize tests', function () {
+    it('behaves as expected', async function () {
+      const cases = [
```
BYK, line 6: Feels like someone needs https://jestjs.io/docs/api#testeachtablename-fn-timeout

chadwhitacre: Oooo, thanks! I searched the docs for "parameters" and came up empty. Is `test` the same as `it`? 🤔

BYK: Yup, they are the same 😊

chadwhitacre: #447 💃

PR #440, opened 2023-05-01
Debug notify-for-triage issue

chadwhitacre:
What do you think of trying this @hubertdeng123?


chadwhitacre:
Not needed! Fixed in https://github.com/getsentry/eng-pipes/pull/441. ☺️

Issue #449, opened 2023-05-04
For triage notifications that come in hourly, only include those in the act fast/overdue queue


hubertdeng123:
It's a bit overwhelming when there are a lot of tickets in the triage queue that are due in over a day

https://sentry.slack.com/archives/G01F3FQ0T41/p1683241196333839

PR #450, opened 2023-05-04
Only display overdue/act fast issues if there are more than 0 of those

hubertdeng123:
closes https://github.com/getsentry/eng-pipes/issues/449

### DIFF
--- a/src/webhooks/pubsub/index.ts
+++ b/src/webhooks/pubsub/index.ts
@@ -360,6 +360,7 @@ export const constructSlackMessage = (
         hasEnoughTimePassedSinceLastNotification &&
         hasEnoughTimePassedSinceIssueCreation;
       const formattedIssues = sortAndFlattenIssuesArray(triageQueueIssues);
+      // Only add issues to triage queue if there ard no overdue or act fast issues
       if (
         triageQueueIssues.length > 0 &&
         overdueIssues.length === 0 &&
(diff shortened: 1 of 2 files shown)


src/webhooks/pubsub/index.ts
```diff
@@ -360,7 +360,11 @@ export const constructSlackMessage = (
         hasEnoughTimePassedSinceLastNotification &&
         hasEnoughTimePassedSinceIssueCreation;
       const formattedIssues = sortAndFlattenIssuesArray(triageQueueIssues);
-      if (triageQueueIssues.length > 0) {
+      if (
+        triageQueueIssues.length > 0 &&
+        overdueIssues.length === 0 &&
+        actFastIssues.length === 0
+      ) {
```
chadwhitacre, line 367: Sorry it's really not obvious to me how this is the fix. 🐭

Approving in the interest of speed but I don't get it.

chadwhitacre (approved):

src/webhooks/pubsub/index.ts
hubertdeng123: Basically, we first populate the `overdueIssues` and `actFactIssues` list if there are any issues that belong to those respective groups. The `triageQueueIssues` is the last one populated. This changes it so we will **only** add to the triage queue if there are no issues that are overdue or close to being overdue.

hubertdeng123: added a comment

chadwhitacre: ty 84a4f795ff133dda8a439592b59288903f6462df

PR #454, opened 2023-05-08
Paginate the check runs and use env vars for test

mattgauntseo-sentry:
This PR fixes an issue where the check runs weren't being paginated meaning eng-pipes treated a large number of the getsentry commits as "fullstack changes" and this PR introduces dotenv to load a new environments file specifically for tests.

This should make it easier to have local env not affect tests.


src/api/github/getChangedStack.ts
```diff
@@ -21,13 +21,13 @@ export async function getChangedStack(
     // We can save on making extra calls to get GH client
     const octokit = client || (await getClient(ClientType.App, OWNER));
 
-    const { data } = await octokit.checks.listForRef({
+    const check_runs = await octokit.paginate(octokit.checks.listForRef, {
       owner: OWNER,
       repo,
       ref,
     });
```
asottile-sentry, line 28: we can probably increase `per_page` to `100` (the maximum github allows) to reduce the pagination a bit

asottile-sentry (approved):

PR #460, opened 2023-05-23
Sync product area labels with GH project containing all issues

hubertdeng123:
This PR adds three main things:
1. Instead of setting up GitHub project automation to do this, eng-pipes bot will add issues to the `Issues Someone Else Cares About` project when GTM, support, or outside Sentry org members create issues.
2. Whenever `Product Area: *` label is added to an issue, it will be added to the `Issues Someone Else Cares About` project if it's not already there. After, the `Product Area: *` label will be synced with the `Product Area` field on the project.
3. Whenever the `Product Area` field changes, the corresponding `Product Area: *` label is added to an issue

Demo:

https://github.com/getsentry/eng-pipes/assets/25517925/f361224d-0ef1-407d-b2d4-d3cda802347b


hubertdeng123:
We'll need to check to make sure the getsantry bot has appropriate permissions for GitHub projects (read and write)

chadwhitacre:
Demo looks great. Does the product area sync go the other way? If you modify the project field does the label change? What happens if you add a second product area label? Are there race conditions?

hubertdeng123:
The product area sync does not go the other way, as far as I could gather there doesn't seem to be anything for GitHub events for project. [Link to GitHub events](https://docs.github.com/en/webhooks-and-events/events/github-event-types)

In a rerouting scenario, if a second product area label is added, the first product area label is removed. The bot will also change the project field to be the product area of the second product area label. Demo below


https://github.com/getsentry/eng-pipes/assets/25517925/76875ba6-30f0-42bf-afbf-3aced704396c

chadwhitacre:
> The product area sync does not go the other way, as far as I could gather there doesn't seem to be anything for GitHub events for project.

Oof, that's rough. Can we reach out to GitHub in our shared Slack channel to see if there are any workarounds besides polling? It feels like this is going to be a pretty common case and we don't want to get out of sync.

chadwhitacre (approved):

hubertdeng123:
actually, looks like there is something I can work with:
https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads?actionType=edited#projects_v2

Going to take some time and investigate this today

hubertdeng123:
>  If you modify the project field does the label change?
Now done

https://github.com/getsentry/eng-pipes/assets/25517925/fa17d22d-2ae2-40ce-9223-69168f2fc5cf

> Are there race conditions?

I did notice one if I switch between Product Area fields extremely fast on the project. It might be due to the delay when applying a label to the issue once a Product Area field value is set. Not sure exactly how to avoid it but it requires a user to change the Product Area field value in ~1 second or less so I don't think we should worry about it too much

package.json
```diff
@@ -43,7 +43,7 @@
     "@octokit/plugin-retry": "^3.0.9",
     "@octokit/rest": "^18.12.0",
     "@octokit/types": "^6.34.0",
-    "@octokit/webhooks": "^9.18.0",
+    "@octokit/webhooks": "^11.0.0",
```
hubertdeng123, line 46: this is needed to subscribe to `projects_v2_item` webhooks (it's new!)

src/brain/issueLabelHandler/route.ts
```diff
@@ -167,7 +173,6 @@ export async function markRouted({
 
   const reasonsToSkip = [
     isNotInARepoWeCareAboutForRouting,
-    isFromABot,
```
hubertdeng123, line 170: this is being removed since we need our bot to be able to route issues if a product area field changes

src/config/index.ts
```diff
@@ -104,6 +105,13 @@ export const WAITING_FOR_PRODUCT_OWNER_LABEL = 'Waiting for: Product Owner';
 export const MAX_TRIAGE_DAYS = 2;
 export const MAX_ROUTE_DAYS = 1;
 
+/**
+ * Issues Someone Else Cares About Project
+ */
+
+export const ISSUES_PROJECT_NODE_ID = "PVT_kwDOABVQ184AOGW8";
+export const PRODUCT_AREA_FIELD_ID = "PVTSSF_lADOABVQ184AOGW8zgJEBno";
```
chadwhitacre, line 113: How are you dev/testing? I know we hard-code other things like Slack channel IDs. Are those overridable somehow for local envs?

hubertdeng123: I've just been replacing the value of these variables. It'd probably make my life easier if I include a way to override these defaults if env variables are defined for dev/testing purposes

hubertdeng123: I can put up a PR for that after this goes in

src/brain/issueLabelHandler/route.ts
```diff
@@ -235,5 +240,16 @@ export async function markRouted({
     body: comment,
   });
 
+  /**
+   * We'll try adding the issue to our global issues project. If it already exists, the existing ID will be returned
+   * https://docs.github.com/en/issues/planning-and-tracking-with-projects/automating-your-project/using-the-api-to-manage-projects#adding-an-item-to-a-project
+   */
+  const issueNodeId: string = await addIssueToProject(payload.issue.node_id, payload.repository.name, payload.issue.number, octokit);
```
chadwhitacre, line 247: Isn't it the `itemId` that is returned here?

src/config/index.ts
chadwhitacre: Can you add a quick commit here? I'm seeing plenty of prior art elsewhere in this file using `process.env`.

src/brain/issueLabelHandler/route.ts
hubertdeng123: good point, changed to `itemId`

src/config/index.ts
hubertdeng123: done

src/brain/issueLabelHandler/route.ts
chadwhitacre: So in other words when a product area project field changes, @getsantry will apply the corresponding product area label (via the new `syncLabelsWithProjectField` handler), and applying the product area label lands us in here, and we want to apply `Status: Untriaged` and `Waiting for: Product Owner`?

src/brain/issueLabelHandler/project.ts
```diff
@@ -0,0 +1,48 @@
...
+    id,
+    payload,
+    ...rest
+}: EmitterWebhookEvent<'projects_v2_item.edited'>) {
+    const tx = Sentry.startTransaction({
+        op: 'brain',
+        name: 'issueLabelHandler.syncLabelsWithProjectField',
+    });
+
+    if (!checkForIssuesProject(payload) || !checkForProductAreaField(payload)|| checkForProjectItemNodeId(payload)) {
```
chadwhitacre, line 31: u no like `shouldSkip` 😏

hubertdeng123: oh woops, I initially did it this way since I think I was refactoring some logic, I can add this back so it stays consistent

src/brain/issueLabelHandler/project.ts
```diff
@@ -0,0 +1,55 @@
...
+    const tx = Sentry.startTransaction({
+        op: 'brain',
+        name: 'issueLabelHandler.syncLabelsWithProjectField',
+    });
+
+
+    const reasonsToDoNothing = [
+        checkForIssuesProject,
+        checkForProductAreaField,
+        checkForProjectItemNodeId
```
chadwhitacre, line 36: Hrm, since we're going the `shouldSkip` route, can we update these names to make it clearer? I think maybe ... ?

`checkForIssuesProject` ⇒ `isNotTheRightProject`
`checkForProductAreaField` ⇒ `isNotTheRightProjectField`
`checkForProjectItemNodeId ` ⇒ `doesNotIncludeANodeID`

src/brain/issueLabelHandler/project.ts
```diff
@@ -0,0 +1,55 @@
...
+    const owner = payload?.organization?.login || "";
+    const octokit = await getClient(ClientType.App, owner);
+    const productArea = await getProductAreaFromProjectField(payload.projects_v2_item.node_id, octokit);
+    const issueInfo = await getIssueDetailsFromNodeId(payload.projects_v2_item.content_node_id, octokit);
+
+    await octokit.issues.addLabels({
+        owner,
+        repo: issueInfo.repo,
+        issue_number: issueInfo.number,
+        labels: [`${PRODUCT_AREA_LABEL_PREFIX}${productArea}`],
```
chadwhitacre, line 51: Won't this maybe be like `PRODUCT_AREA_Issues` or `PRODUCT_AREA_Settings - General`? What's the format of `productArea` here? Is it what we want?

src/brain/issueLabelHandler/project.ts
chadwhitacre: Or maybe `isNotInAProjectWeCareAbout` and `isNotAProjectFieldWeCareAbout` to be parallel with `isNotInARepoWeCareAbout`?

chadwhitacre: When we get to `Waiting for:` labels will we extend this handler to cover that case?

src/brain/issueLabelHandler/project.ts
hubertdeng123: `PRODUCT_AREA_LABEL_PREFIX = "Product Area: "`

So if `productArea` is General or Issues, it will add the label `Product Area: Settings` or `Product Area: General`. The types of product areas can be found on our project [here](https://github.com/orgs/getsentry/projects/76/settings/fields/38012538)

src/brain/issueLabelHandler/project.ts
hubertdeng123: > Or maybe isNotInAProjectWeCareAbout and isNotAProjectFieldWeCareAbout to be parallel with isNotInARepoWeCareAbout?

I like this

> When we get to Waiting for: labels will we extend this handler to cover that case?

Yes, we can extend the handler to apply for the `Status` field too

src/brain/issueLabelHandler/project.ts
hubertdeng123: Behavior should be what we want

hubertdeng123: The only thing here is that we don't currently have a way to sync product area fields on the project with security-as-code like we do with labels

src/brain/issueLabelHandler/route.ts
hubertdeng123: Yes, that's correct

src/brain/issueLabelHandler/index.ts
```diff
@@ -25,4 +26,6 @@ export async function issueLabelHandler() {
   githubEvents.on('issue_comment.created', updateCommunityFollowups);
   githubEvents.removeListener('issues.labeled', ensureOneWaitingForLabel);
   githubEvents.on('issues.labeled', ensureOneWaitingForLabel);
+  githubEvents.removeListener('projects_v2_item.edited', syncLabelsWithProjectField);
+  githubEvents.on('projects_v2_item.edited', syncLabelsWithProjectField);
```
chadwhitacre, line 30: This looks good to me. What do you think of putting the reverse sync in its own handler as well, instead of tightly coupling it to routing? That way we're set up better for `Waiting for` labels.

src/brain/issueLabelHandler/route.ts
chadwhitacre: I feel like that's going to get annoying, it's already annoying when someone changes the product area and does not intend to re-enter the triage queue. I kinda think we should totally decouple the two, `Product Area: Foo` and `Waiting for: Bar` should be always be two operations. Wdyt? And what should we do here?

src/brain/issueLabelHandler/project.ts
chadwhitacre: > Behavior should be what we want

Okay, I'll take your word for it. 😅

> The only thing here is that we don't currently have a way to sync product area fields on the project with security-as-code like we do with labels

Oooh, good point. I guess we should follow up with PRs elsewhere for that, eh?

src/brain/issueLabelHandler/helpers.ts
```diff
@@ -16,7 +23,116 @@ export async function shouldSkip(payload, reasonsToSkip) {
...
+export async function getAllProductAreaNodeIds(octokit: Octokit) {
+  const queryForProductAreaNodeIDs = `query{
+    node(id: "${PRODUCT_AREA_FIELD_ID}") {
+      ... on ProjectV2SingleSelectField {
+        options {
+          id
+          name
+        }
+      }
+    }
```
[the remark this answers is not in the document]
hubertdeng123: Had the same thought here. Tried to find a way around this with something like 
```
gh api graphql -f query='query{
    node(id: "${PRODUCT_AREA_FIELD_ID}") {
      ... on ProjectV2SingleSelectField {
        options(filter: { name: "${PRODUCT_AREA_NAME}" }){
            name,
            id
        }
      }
    }
}'
```

with no luck :/

src/brain/issueLabelHandler/route.ts
hubertdeng123: I'm assuming that the unlabeling/labeling and comment from the bot is what's annoying, and you have a point here. I do want to completely remove the comments from the getsantry bot about when the issue is due by and instead use a field on the project for that timestamp to avoid this kind of noise in the future. Alongside that, when the `Status: Untriaged` labels go away, that will further reduce that noise. 

Do you think it needs to be addressed now? Are you saying that the bot shouldn't care about rerouting if the product area project field changes?

src/brain/issueLabelHandler/index.ts
hubertdeng123: yeah good idea, just separated that into its own handler

src/brain/issueLabelHandler/project.ts
hubertdeng123: yep!

src/brain/issueLabelHandler/helpers.ts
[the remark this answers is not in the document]
hubertdeng123: good idea, posted in #discuss-dev-infra to see if anyone has some input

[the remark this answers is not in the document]
hubertdeng123: no luck :(

https://sentry.slack.com/archives/CTJL7358X/p1685553484946519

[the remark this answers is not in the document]
chadwhitacre: I see. Well, at least we learned about the [explorer](https://docs.github.com/en/graphql/overview/explorer). :)

src/brain/issueLabelHandler/index.ts
chadwhitacre: Ah, hrm. I was actually thinking of factoring out `modifyProjectIssueProductArea` and making it generic, so that it could also be used for "Waiting for" in the future. Could cross that bridge later if you want.

src/brain/issueLabelHandler/route.ts
chadwhitacre: I don't mind the comment, I think it's good communication with the user. The annoying part is when setting a product area automatically retriggers triage, that's not always intended.

src/brain/issueLabelHandler/index.ts
hubertdeng123: Ah got it, I do like having the `projects_v2_item` events in another listener so I think that's a nice change to have when we end up adding more automations in the future. This generic refactor seems trivial enough to implement so I'll do it now.

hubertdeng123: done, made the function generic and named it `modifyProjectIssueField`

src/brain/issueLabelHandler/route.ts
hubertdeng123: Yeah, this is a totally fair point given we've moved away from teams to use product areas. This might be a larger code refactor, do you think this can this be revisited after this PR is in?

chadwhitacre: Yes, agreed, let's ship as-is and revisit down the line.

src/brain/issueLabelHandler/index.ts
chadwhitacre: 08d6d49b5fc5f03a8af818439514271875eea516. Looks good enough for now! :)

src/brain/projectsHandler/project.ts
```diff
@@ -0,0 +1,67 @@
...
+  const reasonsToDoNothing = [
+    isNotInAProjectWeCareAbout,
+    isNotAProjectFieldWeCareAbout,
+    isMissingNodeId,
+  ];
+  if (await shouldSkip(payload, reasonsToDoNothing)) {
+    return;
+  }
+
+  const owner = payload?.organization?.login || '';
```
chadwhitacre, line 48: Coming here from #524 ... I'm wondering why we default to the empty string here. I doubt an `App` octokit loaded with an empty org would work. I think we should let an exception bubble up if `payload.organization.login` does not exist.

PR #465, opened 2023-05-25
Support disabling github metrics via env var

mattgauntseo-sentry:
This hook can be noisy when testing locally.

This is something I end up doing each time I try and run eng-pipes locally.


hubertdeng123 (approved):
Yeah, can definitely be noisy when you don't care about it

PR #478, opened 2023-06-08
Bugfixes for getsantry issue/project handling

hubertdeng123:
1. Only track followups from community when `Waiting for: Community` label is applied
2. When the project field value of an issue is unset, handle gracefully (should resolve https://sentry.sentry.io/issues/4235971037/activity/?project=5246761&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=3&utc=true)

### DIFF
--- a/src/brain/issueLabelHandler/index.test.ts
+++ b/src/brain/issueLabelHandler/index.test.ts
@@ -153,31 +153,6 @@ describe('issueLabelHandler', function () {
     );
   }
 
-  async function editProjectField(
-    projectNodeId?: string,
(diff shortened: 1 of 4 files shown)


src/brain/issueLabelHandler/index.test.ts
```diff
@@ -153,31 +153,6 @@ describe('issueLabelHandler', function () {
     );
   }
 
-  async function editProjectField(
```
hubertdeng123, line 156: unused here

chadwhitacre (approved):

PR #479, opened 2023-06-09
Capture Sentry exceptions when querying from gh graphql api

hubertdeng123:


src/utils/githubEventHelpers.ts
```diff
@@ -63,7 +73,17 @@ export async function getAllProjectFieldNodeIds(projectFieldId: string, octokit:
     }
   }`;
 
-  const response: any = await octokit.graphql(queryForPrjectFieldNodeIDs);
+  let response: any;
+
+  try {
+    response = await octokit.graphql(queryForPrjectFieldNodeIDs);
```
chadwhitacre, line 79:
```suggestion
    response = await octokit.graphql(queryForProjectFieldNodeIDs);
```

hubertdeng123: good catch

chadwhitacre (approved):
It's a little WET to have try/catch everywhere, there could be some wrapper function that does the error handling. It'd be a little complicated though since we want to send progressively more info as we have it, so I'm to land as-is.

hubertdeng123:
> It's a little WET to have try/catch everywhere, there could be some wrapper function that does the error handling.

You're right, let me DRY that up

chadwhitacre:
I like it. 😃

PR #480, opened 2023-06-10
Abstract `Waiting for: *` automations from rerouting

hubertdeng123:
With this change, triaging issues won't happen when issues are routed to new product area unless the issue is `Waiting for: Support`.

### DIFF
--- a/src/brain/issueLabelHandler/route.ts
+++ b/src/brain/issueLabelHandler/route.ts
@@ -16,10 +16,8 @@ import {
   UNKNOWN_LABEL,
   UNROUTED_LABEL,
   UNTRIAGED_LABEL,
-  WAITING_FOR_LABEL_PREFIX,
   WAITING_FOR_PRODUCT_OWNER_LABEL,
   WAITING_FOR_SUPPORT_LABEL,
-  STATUS_FIELD_ID,
 } from '@/config';
 import {
   addIssueToGlobalIssuesProject,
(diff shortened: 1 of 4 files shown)


src/utils/githubEventHelpers.ts
```diff
@@ -48,7 +48,7 @@ export async function addIssueToGlobalIssuesProject(
 
   const response: any = await octokit.graphql(addIssueToGlobalIssuesProjectMutation);
 
-  return response.addProjectV2ItemById.item.id;
+  return response?.addProjectV2ItemById.item.id;
```
hubertdeng123, line 51: I think this resolves https://sentry.sentry.io/issues/4239815868/?project=5246761&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=2&utc=true

src/brain/issueLabelHandler/route.ts
```diff
@@ -113,15 +111,6 @@ export async function markUnrouted({
     body: `Assigning to @${SENTRY_ORG}/support for [routing](https://open.sentry.io/triage/#2-route), due by **<time datetime=${timeToRouteBy}>${readableDueByDate}</time> (${lastOfficeInBusinessHours})**. ⏲️`,
   });
 
-  const itemId: string = await addIssueToGlobalIssuesProject(payload.issue.node_id, repo, issueNumber, octokit);
```
hubertdeng123, line 116: This is duplicate logic, since when `Waiting for: Support` is added, there is automation that kicks in in `followups.ts` that handles this. I have a feeling that's where this error is coming from:
https://sentry.sentry.io/issues/4239815892/?project=5246761&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=3&utc=true

src/brain/issueLabelHandler/route.ts
```diff
@@ -226,9 +216,19 @@ export async function markRouted({
     owner: owner,
     repo: payload.repository.name,
     issue_number: payload.issue.number,
-    labels: [UNTRIAGED_LABEL, WAITING_FOR_PRODUCT_OWNER_LABEL],
+    labels: [UNTRIAGED_LABEL],
   });
 
+  // Only retriage issues if support is routing
+  if (labelNames.includes(WAITING_FOR_SUPPORT_LABEL)) {
```
hubertdeng123, line 223: What do you think about this, `Waiting for: *` label only changes is support is doing the routing?

src/utils/githubEventHelpers.ts
chadwhitacre: 🤞

src/brain/issueLabelHandler/route.ts
chadwhitacre: I like it! This way if some other EPD member at Sentry updates the product area then it will not automatically go to triage, but we can assume Support does mean to retriage. 👍

If you want to use the other suggestion about, then also this:

```suggestion
  if (isBeingRoutedBySupport) {
```

chadwhitacre (approved):
Seems like a good idea. Left a style/clarity suggestion inline.

PR #481, opened 2023-06-13
Move tracking issue due dates to `Issues Someone Else Cares About` project

hubertdeng123:
This moves the due date of responding to issues to `Issues Someone Else Cares About` project instead of getsantry bot comment. 

Why?
The bot shouldn't comment every single time the `Waiting for: *` labels/status change on an issue. That would be way too noisy.

### DIFF
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -1,5 +1,6 @@
 import { EmitterWebhookEvent } from '@octokit/webhooks';
 import * as Sentry from '@sentry/node';
+import moment from 'moment-timezone';
 
 import { ClientType } from '@/api/github/clientType';
 import {
@@ -9,15 +10,21 @@ import {
   WAITING_FOR_COMMUNITY_LABEL,
   WAITING_FOR_LABEL_PREFIX,
   WAITING_FOR_PRODUCT_OWNER_LABEL,
+  RESPONSE_DUE_DATE_FIELD_ID,
+  UNTRIAGED_LABEL,
+  WAITING_FOR_SUPPORT_LABEL,
+  UNROUTED_LABEL,
 } from '@/config';
 import {
   addIssueToGlobalIssuesProject,
   isNotFromAnExternalOrGTMUser,
   modifyProjectIssueField,
   shouldSkip,
+  modifyDueByDate,
 } from '@/utils/githubEventHelpers';
 import { getClient } from '@api/github/getClient';
 import { isFromABot } from '@utils/isFromABot';
+import { calculateSLOViolationTriage, calculateSLOViolationRoute } from '@utils/businessHours';
 
 const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set([
   ...SENTRY_REPOS,
@@ -140,6 +147,7 @@ export async function ensureOneWaitingForLabel({
   const labelToRemove = issue.labels?.find(
     ({ name }) => name.startsWith(WAITING_FOR_LABEL_PREFIX) && name != labelName
   )?.name;
+
   if (labelToRemove != null) {
     await octokit.issues.removeLabel({
       owner: owner,
@@ -157,6 +165,24 @@ export async function ensureOneWaitingForLabel({
   );
 
   await modifyProjectIssueField(itemId, labelName, STATUS_FIELD_ID, octokit);
+
+  let timeToTriageBy
+  if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
+    timeToTriageBy = await calculateSLOViolationTriage(UNTRIAGED_LABEL, issue.labels) || moment().toISOString();
+  }
+  else if (labelName === WAITING_FOR_SUPPORT_LABEL) {
+    timeToTriageBy = await calculateSLOViolationRoute(UNROUTED_LABEL) || moment().toISOString()
+  }
+  else{
+    timeToTriageBy = "";
+  }
+
+  await modifyDueByDate(
+    itemId,
+    timeToTriageBy,
+    RESPONSE_DUE_DATE_FIELD_ID,
+    octokit
+  );
 
   tx.finish();
 }
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -4,27 +4,30 @@ import moment from 'moment-timezone';
 
 import { ClientType } from '@/api/github/clientType';
 import {
+  RESPONSE_DUE_DATE_FIELD_ID,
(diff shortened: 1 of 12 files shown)


src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +166,23 @@ export async function ensureOneWaitingForLabel({
 
   await modifyProjectIssueField(itemId, labelName, STATUS_FIELD_ID, octokit);
 
+  let timeToTriageBy
+  if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
```
hubertdeng123, line 170: move the calculation of due dates to the followups handler that listens for `Waiting for: *` labels and take it out of routing handler

src/utils/githubEventHelpers.ts
```diff
@@ -118,7 +118,39 @@ export async function modifyProjectIssueField(
     projectFieldOption,
     fieldId,
   }
-  await sendQuery(addIssueToGlobalIssuesProjectMutation, data, octokit);
+  await sendQuery(modifyProjectIssueFieldMutation, data, octokit);
+}
+
+export async function modifyDueByDate(
```
hubertdeng123, line 124: why does this need a new function instead of `modifyProjectIssueField`? Its because it's not a single select field option

src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +166,23 @@ export async function ensureOneWaitingForLabel({
 
   await modifyProjectIssueField(itemId, labelName, STATUS_FIELD_ID, octokit);
 
+  let timeToTriageBy
```
chadwhitacre, line 169: Not the best name since it encompasses both triage, routing, and (soon) stale. Eh?

src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +166,23 @@ export async function ensureOneWaitingForLabel({
...
+  if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
+    timeToTriageBy = await calculateSLOViolationTriage(UNTRIAGED_LABEL, issue.labels) || moment().toISOString();
+  }
+  else if (labelName === WAITING_FOR_SUPPORT_LABEL) {
+    timeToTriageBy = await calculateSLOViolationRoute(UNROUTED_LABEL) || moment().toISOString()
+  }
+  else{
+    timeToTriageBy = "";
```
chadwhitacre, line 177: This is where we'll bring in stalebot, eh?

src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +166,23 @@ export async function ensureOneWaitingForLabel({
...
   await modifyProjectIssueField(itemId, labelName, STATUS_FIELD_ID, octokit);
 
+  let timeToTriageBy
+  if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
+    timeToTriageBy = await calculateSLOViolationTriage(UNTRIAGED_LABEL, issue.labels) || moment().toISOString();
+  }
+  else if (labelName === WAITING_FOR_SUPPORT_LABEL) {
+    timeToTriageBy = await calculateSLOViolationRoute(UNROUTED_LABEL) || moment().toISOString()
+  }
+  else{
```
chadwhitacre, line 176: Whitespace. Are we linting? 🤔

src/brain/issueLabelHandler/route.ts
```diff
@@ -101,14 +91,11 @@ export async function markUnrouted({
...
 
-  const timeToRouteBy = await calculateSLOViolationRoute(UNROUTED_LABEL);
-  const { readableDueByDate, lastOfficeInBusinessHours } =
-    await getReadableTimeStamp(timeToRouteBy, UNKNOWN_LABEL);
   await octokit.issues.createComment({
     owner,
     repo: repo,
     issue_number: issueNumber,
-    body: `Assigning to @${SENTRY_ORG}/support for [routing](https://open.sentry.io/triage/#2-route), due by **<time datetime=${timeToRouteBy}>${readableDueByDate}</time> (${lastOfficeInBusinessHours})**. ⏲️`,
+    body: `Assigning to @${SENTRY_ORG}/support for [routing](https://open.sentry.io/triage/#2-route) ⏲️`,
```
chadwhitacre, line 98: I thought the point of this PR was to reduce comment spam. Aren't we still commenting here, just without the due date info? Am I reading this right?

chadwhitacre: Reminds me: how does it show up when we change a field in a project? Does that show up inline on the ticket history at all?

chadwhitacre: I just tested out on https://github.com/getsentry/self-hosted/issues/2205#event-553709b5-b6ef-500f-808e-281bb5f8b55a. I think we get status changes in issue history but not other field changes (I set a value for "Respond By" and am not seeing it in the history).

src/brain/issueLabelHandler/followups.ts
hubertdeng123: you're right :)

let's change that

src/brain/issueLabelHandler/followups.ts
hubertdeng123: Not necessarily needed actually, GH has something neat that includes the `updated_at` property for issues that tells us when any action was last done on an issue. This is just there to indicate there isn't a respond by date, since we're just waiting for community to respond.

See this for the usage of the `updated_at` property
https://github.com/getsentry/eng-pipes/blob/main/src/webhooks/pubsub/stalebot.ts#L82

src/brain/issueLabelHandler/followups.ts
hubertdeng123: looks like I wasn't 🙈

src/brain/issueLabelHandler/route.ts
chadwhitacre: So we'll get a comment every time an issue changes product area but not every time an issue requires a response from that product area. Something like that?

hubertdeng123: > So we'll get a comment every time an issue changes product area but not every time an issue requires a response from that product area. Something like that?

Yep

> I think we get status changes in issue history but not other field changes (I set a value for "Respond By" and am not seeing it in the history).

and yes

chadwhitacre: > a comment every time an issue changes product area

I guess that works for now. We can dial back the comments further if we want in the future.

chadwhitacre (approved):

PR #488, opened 2023-06-16
Move tracking issue due dates to Issues Someone Else Cares About project

hubertdeng123:
a duplicate of https://github.com/getsentry/eng-pipes/pull/481

Need to time this merge better, after slack notifications are updated to use project field

### DIFF
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -1,5 +1,6 @@
 import { EmitterWebhookEvent } from '@octokit/webhooks';
 import * as Sentry from '@sentry/node';
+import moment from 'moment-timezone';
 
 import { ClientType } from '@/api/github/clientType';
 import {
@@ -9,15 +10,21 @@ import {
   WAITING_FOR_COMMUNITY_LABEL,
   WAITING_FOR_LABEL_PREFIX,
   WAITING_FOR_PRODUCT_OWNER_LABEL,
+  RESPONSE_DUE_DATE_FIELD_ID,
+  UNTRIAGED_LABEL,
+  WAITING_FOR_SUPPORT_LABEL,
+  UNROUTED_LABEL,
 } from '@/config';
 import {
   addIssueToGlobalIssuesProject,
   isNotFromAnExternalOrGTMUser,
   modifyProjectIssueField,
   shouldSkip,
+  modifyDueByDate,
 } from '@/utils/githubEventHelpers';
 import { getClient } from '@api/github/getClient';
 import { isFromABot } from '@utils/isFromABot';
+import { calculateSLOViolationTriage, calculateSLOViolationRoute } from '@utils/businessHours';
 
 const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set([
   ...SENTRY_REPOS,
@@ -140,6 +147,7 @@ export async function ensureOneWaitingForLabel({
   const labelToRemove = issue.labels?.find(
     ({ name }) => name.startsWith(WAITING_FOR_LABEL_PREFIX) && name != labelName
   )?.name;
+
   if (labelToRemove != null) {
     await octokit.issues.removeLabel({
       owner: owner,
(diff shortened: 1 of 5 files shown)


src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +169,25 @@ export async function ensureOneWaitingForLabel({
...
+  let timeToRespondBy;
+  if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
+    timeToRespondBy =
+      (await calculateSLOViolationTriage(UNTRIAGED_LABEL, issue.labels)) ||
+      moment().toISOString();
+  } else if (labelName === WAITING_FOR_SUPPORT_LABEL) {
+    timeToRespondBy =
+      (await calculateSLOViolationRoute(UNROUTED_LABEL)) ||
+      moment().toISOString();
+  } else {
```
chadwhitacre, line 181: Should we care about these missing tests?

src/brain/issueLabelHandler/route.ts
```diff
@@ -101,64 +91,33 @@ export async function markUnrouted({
...
     await octokit.teams.getByName({
       org: SENTRY_ORG,
       team_slug: ghTeamSlug,
     }); // expected to throw if team doesn't exist
-    return `Routing to @${SENTRY_ORG}/${ghTeamSlug} for [triage](https://develop.sentry.dev/processing-tickets/#3-triage)`;
+    return `Routing to @${SENTRY_ORG}/${ghTeamSlug} for [triage](https://develop.sentry.dev/processing-tickets/#3-triage) ⏲️`;
   } catch (error) {
     Sentry.captureException(error);
-    return `Failed to route for ${productAreaLabelName}. Defaulting to @${SENTRY_ORG}/open-source for [triage](https://develop.sentry.dev/processing-tickets/#3-triage)`;
+    return `Failed to route for ${productAreaLabelName}. Defaulting to @${SENTRY_ORG}/open-source for [triage](https://develop.sentry.dev/processing-tickets/#3-triage) ⏲️`;
```
chadwhitacre, line 117: I guess we'll update our "route" and "triage" nomenclature later.

src/brain/issueLabelHandler/route.ts
```diff
@@ -181,7 +140,7 @@ export async function markRouted({
   const owner = payload.repository.owner.login;
   const octokit = await getClient(ClientType.App, owner);
   const labelsToRemove: string[] = [];
-  const labelNames = issue?.labels?.map(label => label.name) || [];
+  const labelNames = issue?.labels?.map((label) => label.name) || [];
```
chadwhitacre, line 143: ?

hubertdeng123: doesn't do anything, I can definitely change this back

src/brain/issueLabelHandler/followups.ts
hubertdeng123: wrote some tests to cover those edge cases

src/brain/issueLabelHandler/route.ts
```diff
@@ -67,6 +67,36 @@ function shouldLabelBeRemoved(labelName, target_name) {
   );
 }
 
+async function getReadableTimeStamp(timeToTriageBy, productAreaLabelName) {
```
hubertdeng123, line 70: thought this would be a better place to put this, since it's above all the functions that listen to webhooks

src/utils/githubEventHelpers.ts
```diff
@@ -105,7 +105,7 @@ export async function modifyProjectIssueField(
     octokit
   );
   const singleSelectOptionId = projectFieldNodeIDMapping[projectFieldOption];
-  const addIssueToGlobalIssuesProjectMutation = `mutation {
+  const modifyProjectIssueFieldMutation = `mutation {
```
chadwhitacre, line 108: Fixing a bad name?

hubertdeng123: yessir, this function definitely has nothing to do with adding issues to global issues project

src/brain/issueLabelHandler/route.ts
```diff
@@ -131,36 +162,6 @@ async function routeIssue(octokit, productAreaLabelName) {
...
-      ? 'dddd, MMMM Do [at] HH:mm'
-      : 'dddd, MMMM Do [at] h:mm a';
-  return {
-    readableDueByDate: dueByMoment
-      .tz(OFFICE_TIME_ZONES[lastOfficeInBusinessHours])
-      .format(officeDateFormat),
-    lastOfficeInBusinessHours,
-  };
-}
-
```
chadwhitacre, line 163: Are there material changes to this function? It looks like it's just moving to a different place in the file.

hubertdeng123: see https://github.com/getsentry/eng-pipes/pull/488#discussion_r1237446925 😄

src/brain/issueLabelHandler/followups.ts
```diff
@@ -158,5 +169,25 @@ export async function ensureOneWaitingForLabel({
...
+    timeToRespondBy = '';
+  }
```
chadwhitacre, line 183: Probably not the biggest deal, but aren't we running `calculateSLOViolation*` twice, once for the comment and once for the project field? Won't they be slightly different values when conceptually they should be the same?

hubertdeng123: Yes, for now they may differ by a factor of milliseconds but for our purposes that shouldn't be a big deal since to our logic and we will remove the timestamp from comments soon. I don't see a way where we can fold that logic into one right now since they exist in two different webhook listeners

src/brain/issueLabelHandler/route.ts
chadwhitacre: 😁

chadwhitacre (approved):

PR #489, opened 2023-06-16
Cache the whole client

chadwhitacre:
We've had this comment forever:

> Not sure if we can cache the octokit instance - installation tokens expire
> after an hour, but octokit client may be able to handle this properly.

I found [docs](https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#using-the-octokitjs-sdk-to-authenticate-as-an-app-installation) that say the octokit client _does_ handle this properly!

This is preparatory to adding support for configuring more than one client to support multi-org for #482.

### DIFF
--- a/src/api/github/getClient.ts
+++ b/src/api/github/getClient.ts
@@ -6,7 +6,7 @@ import { GH_USER_TOKEN } from '@/config/index';
 
 import { ClientType } from './clientType';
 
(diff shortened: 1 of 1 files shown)


hubertdeng123 (approved):
looks good!🚀

PR #491, opened 2023-06-16
Factor auth out of _getAppClient

chadwhitacre:
This sets us up to make app auth variable in the org requested.


hubertdeng123 (approved):

src/api/github/getClient.ts
```diff
@@ -63,15 +58,20 @@ export async function getClient(type: ClientType, org: string | null) {
       );
     }
 
+    const auth: AuthInfo = {
```
hubertdeng123, line 61:
```suggestion
    const auth: AppAuthStrategyOptions = {
```

hubertdeng123 (changes requested):

src/api/github/getClient.ts
chadwhitacre: Oof. 2b6603618d230299321d13e3c9a96b404d188623

src/api/github/getClient.ts
```diff
@@ -63,15 +58,20 @@ export async function getClient(type: ClientType, org: string | null) {
...
     if (client === undefined) {
-
       // Bootstrap with a client not bound to an org.
-      const appClient = _getAppClient();
+      const appClient = _getAppClient(auth);
 
       // Use the unbound client to hydrate a client bound to an org.
       const installation = await appClient.apps.getOrgInstallation({ org });
-      client = _getAppClient(installation.data.id);
+      auth.installationId = installation.data.id;
```
chadwhitacre, line 73: I logged locally and also browsed GitHub's UI for existing apps, this does appear to be numeric (per the type def above).

hubertdeng123 (approved):

PR #492, opened 2023-06-16
Use `Waiting for: Product Owner` label to send triage notifications

hubertdeng123:
Uses `Waiting for: Product Owner` instead of `Status: Untriaged` to send slack notifications


src/utils/githubEventHelpers.ts
```diff
@@ -192,6 +192,31 @@ export async function getKeyValueFromProjectField(
...
+      }
+    }`;
+
+  const data = {
+    issueNodeId,
+  };
+  const response = await sendQuery(query, data, octokit);
+
+  return response?.node.fieldValueByName?.text;
+}
```
chadwhitacre, line 218: Lotta uncovered lines here. 🐭

hubertdeng123: You got me 😅, it seems like writing tests here are not super useful, but I guess minimal tests are better than no tests. Will add.

chadwhitacre: > it seems like writing tests here are not super useful

Maybe not, I was just going off of the noise Codecov was making. 😅

src/utils/githubEventHelpers.ts
```diff
@@ -192,6 +192,31 @@ export async function getKeyValueFromProjectField(
...
 
+export async function getIssueDueDateFromProject(
+  issueNodeId: string,
+  octokit: Octokit
+) {
+  const query = `query{
+    node(id: "${issueNodeId}") {
+        ... on ProjectV2Item {
+          id
+          fieldValueByName(name: "Respond By") {
```
chadwhitacre, line 203: Do we have to query by name? If we query by id then it would be immutable, maybe?

chadwhitacre: I guess I like having the flexibility to tinker with the names of the project board and fields without messing up the automation.

hubertdeng123: I also do like that, but I don't think it's possible. Played around with this using the GH graphql explorer and it's only allowing me to input a field value by name
Screenshot 2023-06-21 at 2 43 50 PM

hubertdeng123: This is what using `fieldValues` looks like

```
const query = `query{
    node(id: "${issueNodeId}") {
      ... on ProjectV2Item {
        id
        fieldValues(first: 50) {
          nodes {
            ... on ProjectV2ItemFieldTextValue {
              id
              text
              field {
                ... on ProjectV2Field {
                  id
                  name
                }
              }
            }
          }
        }
      }
    }
  }`;

  const data = {
    issueNodeId,
  };
  const response = await sendQuery(query, data, octokit);
  // When the response due date is empty, the node doesn't exist so we default to empty string
  const issueDueDateInfoNode = response?.node.fieldValues.nodes.find(item => item.field?.id === RESPONSE_DUE_DATE_FIELD_ID) || "";
  return issueDueDateInfoNode.text;
```
Honestly, I really don't like relying on this line here...
```  // When the response due date is empty, the node doesn't exist so we default to empty string
  const issueDueDateInfoNode = response?.node.fieldValues.nodes.find(item => item.field?.id === RESPONSE_DUE_DATE_FIELD_ID) || "";```

Thoughts?

Comment (chadwhitacre) [reply src/utils/githubEventHelpers.ts:203]: > I really don't like relying on this line here

Why not? Because it's inefficient to loop over the field items every time we want to update the due date? Or because there's a possibility of not finding the field? Or ... ?

Comment (chadwhitacre) [reply src/utils/githubEventHelpers.ts:203]: I think it's fine, given what GitHub provides. Ideally they would give us a way to query by ID instead of name on the server side. Since they don't, I think it's okay to do this on our side because the number of fields is almost certainly not going to be that high for this board. It's a very slight inefficiency and we only hit it when we want to update the due date on a ticket, which is not _that_ often (once per comment, basically, so maybe at most once a minute per highly active issue).

Comment (hubertdeng123) [reply src/utils/githubEventHelpers.ts:203]: > Why not? Because it's inefficient to loop over the field items every time we want to update the due date? Or because there's a possibility of not finding the field? Or ... ?

Thing that I don't like is the possibility of not finding the field. I do think it's safe but it seems weird to me that not all items on the project board have the same fields. Will make the change

Comment (chadwhitacre) [reply src/utils/githubEventHelpers.ts:203]: > not all items on the project board have the same fields

Say more about this? Surprises me if true.

Comment (hubertdeng123) [reply src/utils/githubEventHelpers.ts:203]: > Say more about this? Surprises me if true.

Er, I meant to say number of `fieldValues`. Say there are 6 fields for an item on a project board. If the due date field is an empty string, GH graphql api returns 5 nodes for the item.

Comment (chadwhitacre) [reply src/utils/githubEventHelpers.ts]: Ah, gotcha. Yeah, that's weird. :-/

Comment (hubertdeng123) [reply src/utils/githubEventHelpers.ts]: Implemented the changes above. @chadwhitacre are there any other concerns for this to go in?

File: src/utils/githubEventHelpers.ts [hunk 3]
```diff
@@ -192,6 +192,44 @@ export async function getKeyValueFromProjectField(
   return response?.node.fieldValueByName?.name;
 }
 
+export async function getIssueDueDateFromProject(
+  issueNodeId: string,
+  octokit: Octokit
+) {
```
chadwhitacre, line 198:
```suggestion
) {
    // Use fieldValues (and iterate) instead of fieldValuesByName in case the name ever changes
```

chadwhitacre (approved):

src/utils/githubEventHelpers.ts
chadwhitacre: Let's add a comment explaining why we're using fieldValues then I'm good to go. 👍

PR #493, opened 2023-06-20
Update jest version

mattgauntseo-sentry:


chadwhitacre:
Sad CI

mattgauntseo-sentry:
This seems to be causing a number of issues locally and since @hubertdeng123 has fixed the original issue I'm going to close this.

We should still do this though.

PR #494, opened 2023-06-20
Enable VSCode debugger

azaslavsky:
This change saves settings and instructions for easily using the built-in VSCode node.js debugger to step through and break on a running local instance of eng-pipes.


.vscode/launch.json
```diff
@@ -0,0 +1,24 @@
+{
+  // Use IntelliSense to learn about possible attributes.
```
chadwhitacre, line 2: Are comments supported here? Usually don't exist in json afaik. 🐭

azaslavsky: VSCode configuing `.json` files (`launch.json`, `settings.json`, etc) actually use `jsonc`, which is basically JSON that allows single-line comments and trailing commas to make configuration via JSON a little less maddening.

.vscode/tasks.json
```diff
@@ -0,0 +1,21 @@
...
+                "isDefault": true
+            },
+            "presentation": {
+                "reveal": "silent",
+                "panel": "shared"
+            },
+            "problemMatcher": []
+        }
+    ]
+  }
```
hubertdeng123, line 21:
```suggestion
  }
  
```

hubertdeng123 (approved):
nice add!

PR #495, opened 2023-06-20
Bump node-gyp version

hubertdeng123:


yarn.lock
```diff
@@ -6807,22 +6950,23 @@ fsevents@~2.1.2:
   linkType: hard
 
 "node-gyp@npm:latest":
-  version: 7.1.2
-  resolution: "node-gyp@npm:7.1.2"
+  version: 9.4.0
```
hubertdeng123, line 6953: look here

chadwhitacre: Needle in a haystack!

chadwhitacre: There's no way to keep the diff smaller?

mattgauntseo-sentry (approved):
LGTM - It seems like fsevents wants node-gyp latest, so this should be safe enough.

yarn.lock
hubertdeng123: I don't think so :(

I think this should be safe enough though

PR #496, opened 2023-06-20
Lint files

hubertdeng123:


chadwhitacre (approved):

chadwhitacre:
How can we ensure we stay linted and don't have to do a big catch-up again later?

chadwhitacre:
> How can we ensure we stay linted and don't have to do a big catch-up again later?

Ah, seeing https://github.com/getsentry/eng-pipes/issues/497 now. 👍

PR #499, opened 2023-06-21
Fix numbering bug

azaslavsky:
The bug appeared because numbers were assigned to the output text before
sorting occurred. In addition to some DRY cleanup, we now create a
`SlackMessageUnorderedIssueItem` which stores all of the output strings
we'll need without numbering. Once we sort, we then map over the
now-sorted list to map all of our `SlackMessageUnorderedIssueItem`s to
`SlackMessageOrderedIssueItem`s via the new
`addOrderingToSlackMessageItem` helper function.

### DIFF
--- a/src/webhooks/pubsub/slackNotifications.ts
+++ b/src/webhooks/pubsub/slackNotifications.ts
@@ -13,7 +13,15 @@ const GH_API_PER_PAGE = 100;
 const DEFAULT_PRODUCT_AREA_LABEL = 'Product Area: Other';
 const getChannelLastNotifiedTable = () => db('channel_last_notified');
 
-type SlackMessageIssueItem = {
+// An item with all of its relevant properties stringified as markdown.
+type SlackMessageUnorderedIssueItem = {
+  triageBy: string;
+  issueLink: string;
+  timeLeft: string;
+}
+
+// Like the above, but the numerical prefix to each row now has a correct numerical ordering.
+type SlackMessageOrderedIssueItem = {
   triageBy: string;
   fields: [
     // Issue title and link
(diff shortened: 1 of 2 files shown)


src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -74,6 +82,30 @@ const getIssueProductAreaLabel = (issue: Issue) => {
   return getLabelName(label) || DEFAULT_PRODUCT_AREA_LABEL;
 };
 
+// Note that the `ordinal` field is the literal number that will show up, not the index in the
+// owning array. For example, it is it the caller's responsibility to offset the 0-indexed entries
```
chadwhitacre, line 86:
```suggestion
// owning array. For example, it is the caller's responsibility to offset the 0-indexed entries
```

src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -89,7 +121,7 @@ export const getTriageSLOTimestamp = async (
   const routingEvents = issues.filter(
     (event) =>
       // @ts-ignore - We _know_ a `label` property exists on `labeled` events
-      event.user.type === 'Bot' && event.user.login === 'getsantry[bot]'
+      event.user.type === 'Bot' // && event.user.login === 'getsantry[bot]'
```
chadwhitacre, line 124: ?

src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -113,12 +145,13 @@ export const constructSlackMessage = (
   productAreaToIssuesMap: Record<string, IssueSLOInfo[]>,
   now: moment.Moment
 ) => {
+  console.log(productAreaToIssuesMap);
```
chadwhitacre, line 148: Maybe you're not ready for review yet. 😬

src/webhooks/pubsub/slackNotifications.ts
azaslavsky: Done.

src/webhooks/pubsub/slackNotifications.ts
azaslavsky: Whoops, some dev slipped in. Fixed!

src/webhooks/pubsub/slackNotifications.ts
azaslavsky: Removed.

src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -235,58 +229,41 @@ export const constructSlackMessage = (
...
+  
       const sortAndFlattenIssuesArray = (issues) =>
         issues
           .sort(
             (a, b) =>
               moment(a.triageBy).valueOf() - moment(b.triageBy).valueOf()
           )
-          .map((item) => item.fields)
+          .map((item, index) => {
+            return addOrderingToSlackMessageItem(item, index + 1).fields;
```
hubertdeng123, line 263: Wondering what this behavior would look like if there are issues in both the overdue and act fast queue. Will it reset the numbering for each respective queue?

Rough Example:

Overdue 
1. Issue One
2. Issue Two

Act Fast
1. Issue Three
2. Issue Four

azaslavsky: I've added another test that covers this scenario in the latest patcheset.

chadwhitacre (approved):

PR #500, opened 2023-06-22
Be more explicit about dry run flag

hubertdeng123:
This is failing tests when modifying `DRY_RUN` locally with any string value. Tests will only pass if `DRY_RUN` is undefined

Also added some other common env variables to `.env.test` to preserve those variables for testing so env variables configured with dev won't fail tests.


src/utils/metrics.ts
```diff
@@ -170,7 +170,8 @@ export async function _insert(
   data: Record<string, any>,
   targetConfig: TargetConfig
 ) {
-  if (process.env.DRY_RUN) {
+  // DRY_RUN ready through env variable will be a string, so we need to be more explicit here.
```
chadwhitacre, line 173: Typo:

```suggestion
  // DRY_RUN read through env variable will be a string, so we need to be more explicit here.
```

But also ... really? Empty string is falsey. We're getting something other than that here? 🤔

```
$ node                                                                                                         
Welcome to Node.js v19.9.0.
Type ".help" for more information.
> Boolean('')
false
> Boolean(' ')
true
> 
```

hubertdeng123: Well, I was using `DRY_RUN=0` which wasn't working for me

chadwhitacre: Ah, well, yeah, that makes sense.  :)

If I comment out `DRY_RUN` (don't set it at all) then the test suite passes:

```
Test Suites: 43 passed, 43 total
Tests:       1 skipped, 297 passed, 298 total
Snapshots:   29 passed, 29 total
Time:        18.966 s, estimated 19 s
```

hubertdeng123: I think it doesn't hurt to be more explicit here. Also, adding `DRY_RUN=` to the `.env.test` seems to make the tests pass even if I set `DRY_RUN=1` in my `.env` file

chadwhitacre: > Also, adding `DRY_RUN=` to the `.env.test` seems to make the tests pass even if I set `DRY_RUN=1` in my `.env` file

Makes sense, I guess it's empty string in that case. Nice find.

chadwhitacre: > I think it doesn't hurt to be more explicit here.

Okay but in that case we should move the parsing to `src/config/index.ts` and use that from `metrics.ts`, and we should also support "yes" and we should be case insensitive, and are we doing similar parsing for any other boolean-ish envvars?

hubertdeng123: Yeah, that does sound reasonable. Scanning through I see one example of another boolean env var:
```
export const DISABLE_GITHUB_METRICS =
  process.env.DISABLE_GITHUB_METRICS == 'true';
```

chadwhitacre: > I see one example of another boolean env var

How about we make `DRY_RUN` behave the same as that for consistency with the status quo?

hubertdeng123: > How about we make DRY_RUN behave the same as that for consistency with the status quo?

Sounds good to me.

src/config/index.ts
```diff
@@ -30,6 +30,10 @@ export const TEAM_OSPO_CHANNEL_ID = // #team-ospo
   process.env.TEAM_OSPO_CHANNEL_ID || 'G01F3FQ0T41';
 export const DISABLE_GITHUB_METRICS =
   process.env.DISABLE_GITHUB_METRICS == 'true';
```
chadwhitacre, line 32: Support '1' here as well? 🤷

src/config/index.ts
```diff
@@ -30,6 +30,10 @@ export const TEAM_OSPO_CHANNEL_ID = // #team-ospo
...
+export const DRY_RUN =
+  process.env.DRY_RUN == 'true' || process.env.DRY_RUN == '1';
```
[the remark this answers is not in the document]
hubertdeng123: I don't like supporting `yes` here as it's not your typical Boolean value. I think it's fine to put `process.env.DRY_RUN == 'true' || process.env.DRY_RUN == '1';` instead of complicating things. In Javascript, `true` and `false` are booleans, not `TRUE` or `FALSE` and so on so this seems fine to do

src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -27,7 +27,7 @@ type SlackMessageUnorderedIssueItem = {
   triageBy: string;
   issueLink: string;
   timeLeft: string;
-}
+};
```
azaslavsky, line 30: Thanks for linting this! :)

src/brain/githubMetrics/index.test.ts
```diff
@@ -76,10 +81,10 @@ describe('github webhook', function () {
   });
 
   it('does not call insert on dry run', async function () {
```
azaslavsky, line 83: nit: maybe worth doing the test again for 1/0 rather than true/false, since we now support both variants?

.env.test
```diff
@@ -29,3 +29,6 @@ FEED_DEV_INFRA_CHANNEL_ID="13141516"
 DISABLE_GITHUB_METRICS="false"
 FORCE_USER_TOKEN_GITHUB_CLIENT=
 FORCE_GET_USER_BY_SLACK_ID=
+TEAM_OSPO_CHANNEL_ID=
```
azaslavsky, line 32: What do this line and the one below it mean? Are they related to DRY_RUN, or separate things that are being added to avoid a whole separate PR? If the latter, it might be useful to add a line in the commit body that you've also added this orthogonal stuff to the example env.

azaslavsky (approved):

.env.test
hubertdeng123: Yeah, they're separate things I noticed to avoid a whole separate PR since it's touching on the same thing. I'll add to commit body

src/brain/githubMetrics/index.test.ts
hubertdeng123: I actually couldn't get that to work since I believe there's something finnicky around env variables being read into the config once during initialization. I spent a while trying to get it to work then gave up so decided not to go down that route since it's not superrr important

PR #503, opened 2023-06-30
Refactor auth option parsing out of getClient

chadwhitacre:
Part of https://github.com/getsentry/eng-pipes/issues/482, after #509, before #504.

### DIFF
--- a/src/api/github/getClient.ts
+++ b/src/api/github/getClient.ts
@@ -1,18 +1,12 @@
 import { createAppAuth } from '@octokit/auth-app';
 
 import { GH_USER_TOKEN } from '@/config/index';
+import { AppAuthStrategyOptions, GH_APP_AUTH_OPTIONS } from '@/config/index';
 
 import { ClientType } from './clientType';
 import { OctokitWithRetries } from './octokitWithRetries';
 
 const _CLIENTS_BY_ORG = new Map();
-
-interface AppAuthStrategyOptions {
-  // I didn't find something great in @octokit/types.
-  appId: number;
-  privateKey: string;
-  installationId?: number;
-}
 
 function _getUserClient() {
   return new OctokitWithRetries({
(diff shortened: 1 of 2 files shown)


src/api/github/getClient.ts
```diff
@@ -46,22 +40,18 @@ export async function getClient(type: ClientType, org: string | null) {
...
       );
     }
 
-    const auth: AppAuthStrategyOptions = {
-      appId: Number(process.env.GH_APP_IDENTIFIER),
-      privateKey: process.env.GH_APP_SECRET_KEY?.replace(/\\n/g, '\n'),
-    };
+    const auth = GH_APP_AUTH_OPTIONS.get('__tmp_org_placeholder__');
+    if (auth === undefined) {
+      throw new Error(
```
azaslavsky, line 51: Maybe worthwhile having a test that covers this error? Or at least leaving a TODO for the future?

src/config/index.ts
```diff
@@ -166,6 +166,35 @@ export const RESPONSE_DUE_DATE_FIELD_ID =
  */
 export const GH_USER_TOKEN = process.env.GH_USER_TOKEN || '';
 
+/**
+ * App auth strategy options. For branding purposes, we have one app per org
+ * (getsantry for getsentry, covecod for codecov, &c). Down in getClient we
```
azaslavsky, line 171: nit: "&c" -> "etc"?

src/config/index.ts
```diff
@@ -166,6 +166,35 @@ export const RESPONSE_DUE_DATE_FIELD_ID =
...
+export interface AppAuthStrategyOptions {
+  // I didn't find something great in @octokit/types.
+  appId: number;
+  privateKey: string;
+  installationId?: number;
+}
+
+export const GH_APP_AUTH_OPTIONS = new Map<string, AppAuthStrategyOptions>();
+
+if (process.env.GH_APP_IDENTIFIER && process.env.GH_APP_SECRET_KEY) {
```
azaslavsky, line 191: What happens if one or both of these aren't set? Should we bubble some kind of error up, or is this situation dealt with later when we do operations that require AUTH? If we assume that these must always be set for the GH integration to function properly, may be clearer to put an `else` here and error closer to the source of the problem.

azaslavsky (changes requested):
LGTM modulo comments!

src/config/index.ts
chadwhitacre: This change preserves the status quo, which is that GitHub auth errors aren't discovered until we try to use the client (inside of `getClient`, when we use an unbound client to find an installation ID). I figure this is probably fine since Dev Infra or someone else may want to be able to run `eng-pipes` without configuring GitHub.

src/config/index.ts
chadwhitacre: Bytes are expensive. ;-)

Done in ffb23786d62d39048324d6ac40efd90ad9f368b7.

src/api/github/getClient.ts
chadwhitacre: https://github.com/getsentry/eng-pipes/pull/509

chadwhitacre: Chaining this PR on that one. 👍

hubertdeng123 (approved):

azaslavsky (approved):

PR #504, opened 2023-06-30
Tidy up getClient usage

chadwhitacre:
Part of #482, after #503, before #505.

Noticed these in https://github.com/getsentry/eng-pipes/issues/482#issuecomment-1613944724. By getting rid of some of this noise around `getClient`, it should hopefully be a bit easier to tackle the `owner`/`OWNER` discrepancy, which is the primary blocker for https://github.com/getsentry/eng-pipes/issues/482 atm.

### DIFF
--- a/src/api/github/getChangedStack.ts
+++ b/src/api/github/getChangedStack.ts
@@ -12,14 +12,9 @@ const BACKEND_CHANGE_CHECK_NAME = 'only backend changes';
  * Given a commit and repo, check GitHub Checks to see what type of changes
  * were made in the commit (frontend, backend, fullstack)
  */
-export async function getChangedStack(
-  ref: string,
-  repo: string,
-  client?: Octokit
-) {
+export async function getChangedStack(ref: string, repo: string) {
   try {
-    // We can save on making extra calls to get GH client
-    const octokit = client || (await getClient(ClientType.App, OWNER));
+    const octokit = await getClient(ClientType.App, OWNER);
 
     const check_runs = await octokit.paginate(octokit.checks.listForRef, {
       owner: OWNER,
--- a/src/api/github/getChangedStack.ts
+++ b/src/api/github/getChangedStack.ts
@@ -1,4 +1,3 @@
-import { Octokit } from '@octokit/rest';
 import * as Sentry from '@sentry/node';
 
 import { ClientType } from '@/api/github/clientType';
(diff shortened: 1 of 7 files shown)


src/utils/getOssUserType.ts
```diff
@@ -54,6 +53,8 @@ export async function getOssUserType(
   let status: number | null;
   let check: 'Org' | 'Team';
 
+  const octokit = await getClient(ClientType.User);
```
chadwhitacre, line 56: The second arg is dropped on the floor in the `ClientType.User` case.

https://github.com/getsentry/eng-pipes/blob/3de4f3a83da3a2a96eed03ce526492e94ba78ed0/src/api/github/getClient.ts#L36-L41

src/api/github/getSentryPullRequestsForGetsentryRange.test.ts
```diff
@@ -5,26 +5,26 @@ import { getClient } from '@api/github/getClient';
 import { getSentryPullRequestsForGetsentryRange } from './getSentryPullRequestsForGetsentryRange';
 
 describe('getSentryPullRequestsForGetsentryRange', function () {
-  let getsentry;
+  let octokit;
```
chadwhitacre, line 8: The rest of our tests use `octokit`.

src/api/github/getChangedStack.ts
```diff
@@ -12,14 +12,9 @@ const BACKEND_CHANGE_CHECK_NAME = 'only backend changes';
  * Given a commit and repo, check GitHub Checks to see what type of changes
  * were made in the commit (frontend, backend, fullstack)
  */
-export async function getChangedStack(
-  ref: string,
-  repo: string,
-  client?: Octokit
-) {
+export async function getChangedStack(ref: string, repo: string) {
```
chadwhitacre, line 15: I find two calls to this function and neither uses the `client` arg. Even so, with #489 there's no material performance benefit to avoiding getClient.

src/api/github/getRelevantCommit.ts
```diff
@@ -12,10 +12,9 @@ import { getClient } from '@api/github/getClient';
  * caused a bump commit in getsentry (whose commit message references the sentry commit).
  * In this case, it will return the sentry commit.
  */
-export async function getRelevantCommit(ref: string, client?: Octokit) {
```
chadwhitacre, line 15: As with `getChangedStack`, we don't realize any performance gain from passing an octokit around vs. calling getClient to get the same object from the (new) cache. Here though we were actually using the parameter in two of the four calls, so that is undone.

src/api/github/getChangedStack.ts
mattgauntseo-sentry: Yeah I copy pasta'd a lot of stuff from freight, so no concerns from me.

mattgauntseo-sentry (approved):

hubertdeng123 (approved):
LGTM, given that we only really care about TTR logic for codecov

PR #507, opened 2023-07-05
Add team-engineering alerts

mattgauntseo-sentry:


joshuarli (approved):

src/brain/gocdSlackFeeds/index.ts
```diff
@@ -40,8 +45,32 @@ const devinfraFeed = new DeployFeed({
...
+  channelID: FEED_ENGINEERING_CHANNEL_ID,
+  msgType: SlackMessage.FEED_ENGINGEERING_DEPLOY,
+  pipelineFilter: (pipeline) => {
+    if (!ENGINEERING_PIPELINE_FILTER.includes(pipeline.name)) {
+      return false;
+    }
+
+    if (pipeline.stage.name.toLowerCase().indexOf('deploy') === -1) {
+      return false;
+    }
```
IanWoodard, line 60: In what case would the stage name not be deploy? Also, should we also opt to use `includes` here rather than `indexOf`?

joshuarli (approved):

src/brain/gocdSlackFeeds/index.ts
IanWoodard: Couldn't we combine these two if statements? I understand possibly wanting to keep them separate for better readability though, up to you!

IanWoodard (approved):

src/brain/gocdSlackFeeds/index.ts
mattgauntseo-sentry: Going to keep separate in the name of readability.

Stage names you can see here: https://deploy.getsentry.net/go/pipeline/activity/getsentry-backend

And +1 for includes.

PR #512, opened 2023-07-08
Rename OWNER to GETSENTRY_ORG

chadwhitacre:
Part of #482, after #511, before #513.

This makes it clearer what the knob means, and it fits with `GETSENTRY_REPO` and `SENTRY_REPO`. I expect I'll add a `CODECOV_ORG` and utilize that in pubsub and potentially elsewhere.

This also folds the `SENTRY_ORG` knob into `GETSENTRY_ORG` since it was the same as `OWNER` (but `OWNER` was much more popular).


hubertdeng123 (approved):
makes complete sense 😁

PR #516, opened 2023-07-10
Remove fallback to determining due date by issue comment

hubertdeng123:
deprecated logic, needs to be removed


chadwhitacre:
You're confident we don't have any more of these kinds of comments out in the wild? Or that if we do we're not using them anymore (all issues that had these now also have timestamps on the project board)?

hubertdeng123:
> You're confident we don't have any more of these kinds of comments out in the wild?

Yes, and if we do they're either closed or appropriately addressed (resolved or backlogged). Otherwise they would appear on our [401 dashboard](https://sentryio.cloud.looker.com/dashboards/401)

chadwhitacre (approved):

PR #517, opened 2023-07-10
Remove usage of unrouted and untriaged labels

hubertdeng123:
👋


src/utils/businessHours.test.ts
```diff
@@ -279,7 +279,9 @@ describe('businessHours tests', function () {
     });
 
     it('should calculate SLO violation if label is unrouted', async function () {
-      const result = await calculateSLOViolationRoute(UNROUTED_LABEL);
+      const result = await calculateSLOViolationRoute(
```
azaslavsky, line 282: nit: Not strictly in this PR, but still confusing: on L262, I found it pretty hard to parse the double negative "not unrouted". Maybe just "if label is routed"?

azaslavsky (approved):

chadwhitacre (approved):
This PR seems to do a fine job of search/replacing `Status` for `Waiting for` labels. I think there's some deeper refactoring that could go along with this, but no need to block on that here.

PR #519, opened 2023-07-11
Factor out a function to load app config

chadwhitacre:
Part of #482.

Since we're going to have somewhat complex logic to parse envvars into app config, we should test as we go. Factoring out a function gives us something to test.


src/api/github/appAuthStrategyOptions.ts
```diff
@@ -0,0 +1,7 @@
+// I didn't find something great in @octokit/types.
+
+export interface AppAuthStrategyOptions {
```
hubertdeng123, line 3: should this belong in `src/types` instead?

src/config/loadGitHubApps.ts
```diff
@@ -0,0 +1,14 @@
+import { AppAuthStrategyOptions } from '@api/github/appAuthStrategyOptions';
+
+export function loadGitHubApps(env) {
```
hubertdeng123, line 3: does this function need to pass in environment variables as `env`? Can we just use `process.env` here?

hubertdeng123 (changes requested):

src/config/loadGitHubApps.ts
chadwhitacre: It makes testing much easier.

src/api/github/appAuthStrategyOptions.ts
chadwhitacre: Oooh, good call! Done in b98db3a7bea4da859f66107157f227636c120bac.

src/config/loadGitHubApps.ts
hubertdeng123: In that case, can we just pass in `GH_APP_IDENTIFIER` and `GH_APP_SECRET_KEY` instead? feels weird to me to pass all the env variables just to use two of them

chadwhitacre: Filtering out unused envvars is part of the logic we want to test, no? Does da690d21b0e35db187ab60f147291a4d9e40a834 help? It changes the function name to `loadGitHubAppsFromEnvironment` and adds a few more tests.

chadwhitacre: Also keep in mind that the next PR is going to add multi-org support, for which we're going to need to iterate over the whole process.env to pull out the vars we want.

hubertdeng123 (approved):

PR #521, opened 2023-07-12
Scope project board config to app (org)

chadwhitacre:
Part of #482, behind #520.

This is necessary since different orgs have different "GitHub Issues Someone Else Cares About" project boards. There's a lot more refactoring that could happen here (like putting the `octokit` client on each `GitHubApp`) but this PR is big enough as it is.

### DIFF
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -31,20 +31,20 @@ const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set([
   ...SENTRY_MONOREPOS,
 ]);
 
-function isNotInARepoWeCareAboutForFollowups(app, payload) {
+function isNotInARepoWeCareAboutForFollowups(payload) {
   return !REPOS_TO_TRACK_FOR_FOLLOWUPS.has(payload.repository.name);
 }
 
-function isNotWaitingForLabel(app, payload) {
+function isNotWaitingForLabel(payload) {
   return !payload.label?.name.startsWith(WAITING_FOR_LABEL_PREFIX);
 }
 
-function isContractor(app, payload) {
+function isContractor(payload) {
   // Contractors are outside collaborators on GitHub
   return payload.comment.author_association === 'COLLABORATOR';
 }
 
-function isNotWaitingForCommunity(app, payload) {
+function isNotWaitingForCommunity(payload) {
   const { issue } = payload;
   return !issue?.labels.some(
     ({ name }) => name === WAITING_FOR_COMMUNITY_LABEL
@@ -73,7 +73,7 @@ export async function updateCommunityFollowups({
     isFromABot,
   ];
 
-  if (await shouldSkip(app, payload, reasonsToDoNothing)) {
+  if (await shouldSkip(payload, app, reasonsToDoNothing)) {
     return;
   }
 
@@ -137,7 +137,7 @@ export async function ensureOneWaitingForLabel({
     isNotInARepoWeCareAboutForFollowups,
     isNotWaitingForLabel,
   ];
-  if (await shouldSkip(app, payload, reasonsToDoNothing)) {
+  if (await shouldSkip(payload, app, reasonsToDoNothing)) {
     return;
   }
 
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -2,29 +2,28 @@ import { EmitterWebhookEvent } from '@octokit/webhooks';
 import * as Sentry from '@sentry/node';
 import moment from 'moment-timezone';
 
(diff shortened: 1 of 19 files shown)


src/brain/issueLabelHandler/followups.ts
```diff
@@ -2,50 +2,49 @@ import { EmitterWebhookEvent } from '@octokit/webhooks';
...
+} from '@utils/githubEventHelpers';
 import { isFromABot } from '@utils/isFromABot';
 
 const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set([
   ...SENTRY_REPOS,
   ...SENTRY_MONOREPOS,
 ]);
 
-function isNotInARepoWeCareAboutForFollowups(payload) {
+function isNotInARepoWeCareAboutForFollowups(app, payload) {
```
chadwhitacre, line 34: This is a lot of the change here, adding `app` to all of the `shouldSkip`s, because a few of them need it. Now that I think about it, it might've been better to make that the second arg, that way if it's not needed it doesn't matter. 😬 😭

src/brain/issueLabelHandler/followups.ts
```diff
@@ -102,16 +103,18 @@ export async function updateCommunityFollowups({
...
     issueNumber,
     octokit
   );
 
   await modifyProjectIssueField(
+    app,
     itemId,
     WAITING_FOR_PRODUCT_OWNER_LABEL,
-    STATUS_FIELD_ID,
+    app.project.status_field_id,
```
chadwhitacre, line 117: Here's another big thing throughout, taking config from the `app` vs. the global config vars.

src/config/index.ts
```diff
@@ -154,19 +157,6 @@ export const SENTRY_REPOS = [
...
-
-export const ISSUES_PROJECT_NODE_ID =
-  process.env.ISSUES_PROJECT_NODE_ID || 'PVT_kwDOABVQ184AOGW8';
-export const PRODUCT_AREA_FIELD_ID =
-  process.env.PRODUCT_AREA_FIELD_ID || 'PVTSSF_lADOABVQ184AOGW8zgJEBno';
-export const STATUS_FIELD_ID =
-  process.env.STATUS_FIELD_ID || 'PVTSSF_lADOABVQ184AOGW8zgI_7g0';
-export const RESPONSE_DUE_DATE_FIELD_ID =
-  process.env.RESPONSE_DUE_DATE_FIELD_ID || 'PVTF_lADOABVQ184AOGW8zgLLxGg';
-
```
chadwhitacre, line 169: Yoink!

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre, line 1: Here's the heart of the change, a couple classes `GitHubApp` and `GitHubAppRegistry`. We have one registry globally and we populate it with multiple apps, that we can load from where-ever based on org slug.

src/utils/getOssUserType.ts
```diff
@@ -19,7 +19,7 @@ const _USER_CACHE = new Map<string, CachedUser>();
 export async function getOssUserType(
   payload: Record<string, any>
 ): Promise<UserType | null> {
-  if (isFromABot(payload)) {
+  if (isFromABot(null /* no app here */, payload)) {
```
chadwhitacre, line 22: Great place to have put `app` second. 🙄

src/brain/issueLabelHandler/followups.ts
hubertdeng123: agreed here, optional parameter seems to make sense instead of passing unused ones.

hubertdeng123 (changes requested):
Main concern is [here](https://github.com/getsentry/eng-pipes/pull/521/files#r1260408957), otherwise this looks good to me 👍

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,13 +1,68 @@
...
+  auth: AppAuthStrategyOptions;
+  project;
+
+  constructor(org: string, opts) {
+    this.org = org;
+    this.auth = {
+      appId: Number(opts.GH_APP_IDENTIFIER),
+      privateKey: opts.GH_APP_SECRET_KEY.replace(/\\n/g, '\n'),
+    };
+    this.project = {
```
hubertdeng123, line 14: guessing for codecov an if conditional will be introduced here?

chadwhitacre: I was thinking of moving the defaults out of our code and into envvars.

hubertdeng123 (approved):

src/brain/issueLabelHandler/followups.ts
chadwhitacre: Done in 1793a82662a507af391475f947c5386ed76ecc0d.

PR #522, opened 2023-07-13
Use more appropriate names for groups of sentry repos

hubertdeng123:


src/webhooks/pubsub/stalebot.ts
```diff
@@ -3,14 +3,17 @@ import moment from 'moment-timezone';
...
+  SENTRY_REPOS_WITHOUT_ROUTING,
   STALE_LABEL,
   WAITING_FOR_COMMUNITY_LABEL,
 } from '@/config';
 import { GitHubApp } from '@/config/loadGitHubAppsFromEnvironment';
 
-const REPOS_TO_TRACK_FOR_STALEBOT = [...SENTRY_MONOREPOS, ...SENTRY_SDK_REPOS];
+const REPOS_TO_TRACK_FOR_STALEBOT = [
+  ...SENTRY_REPOS_WITH_ROUTING,
+  ...SENTRY_REPOS_WITHOUT_ROUTING,
```
chadwhitacre, line 15: How about making a `SENTRY_REPOS` in `config` since we do this combining more than once?

hubertdeng123: done.

src/brain/issueLabelHandler/followups.ts
```diff
@@ -26,10 +25,7 @@ import {
 } from '@utils/githubEventHelpers';
 import { isFromABot } from '@utils/isFromABot';
 
-const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set([
-  ...SENTRY_SDK_REPOS,
-  ...SENTRY_MONOREPOS,
-]);
+const REPOS_TO_TRACK_FOR_FOLLOWUPS = new Set(SENTRY_REPOS);
```
chadwhitacre, line 28: Is there a good reason to keep `REPOS_TO_TRACK_FOR_FOLLOWUPS` around vs. referencing `SENTRY_REPOS` directly like we do for the stalebot case?

src/brain/issueLabelHandler/route.ts
```diff
@@ -20,7 +20,7 @@ import {
 } from '@utils/githubEventHelpers';
 import { slugizeProductArea } from '@utils/slugizeProductArea';
 
-const REPOS_TO_TRACK_FOR_ROUTING = new Set(SENTRY_MONOREPOS);
+const REPOS_TO_TRACK_FOR_ROUTING = new Set(SENTRY_REPOS_WITH_ROUTING);
```
chadwhitacre, line 23: Sticks out here especially, since the names are basically the same. 😂

chadwhitacre (approved):
Looks good with minor suggestions.

src/brain/issueLabelHandler/followups.ts
hubertdeng123: I think it's mostly for the performance, looking up an item in a set is O(1) vs in a list which is O(n)

hubertdeng123: And since this is declared at the top, the set is only created once. So it would need to be here so it's a bit more performant

azaslavsky: Is there anywhere where this will matter in practice? For anything we're practically doing, I imagine a list of like 40 (<- is that right in order of magnitude terms) items is about as fast to scan as a similarly sized set. I do find longer, more descriptive names better for maintainability - maybe we just pick one of these (I vote `REPOS_TO_TRACK_FOR_FOLLOWUPS`) and keep that as the name everywhere?

azaslavsky (approved):

src/brain/issueLabelHandler/followups.ts
hubertdeng123: No harm in declaring `SENTRY_REPOS` as a set actually in config, I'll do that. Unfortunately `REPOS_TO_TRACK_FOR_FOLLOWUPS` won't work because we use the same set of repos for determining which repos to enable stalebot for

PR #523, opened 2023-07-13
Start teaching the app loader about envvar numbers

chadwhitacre:
Part of #482, before #524.

We need two passes to parsing, one where we collect configs based on the number in envvars of the form `GH_APP_1_FOO`, and then another where we organize apps by org slug. This PR stubs out this two-pass structure without yet switching to the new envvars.

### DIFF
--- a/src/config/loadGitHubAppsFromEnvironment.test.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.test.ts
@@ -7,6 +7,7 @@ describe('loadGitHubAppsFromEnvironment ', function () {
         [
           '__tmp_org_placeholder__',
           {
+            num: 1,
             org: '__tmp_org_placeholder__',
             auth: {
               appId: 42,
@@ -59,6 +60,7 @@ describe('loadGitHubAppsFromEnvironment ', function () {
         [
           '__tmp_org_placeholder__',
           {
+            num: 1,
             org: '__tmp_org_placeholder__',
             auth: {
               appId: 42,
--- a/src/config/loadGitHubAppsFromEnvironment.test.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.test.ts
@@ -42,7 +42,7 @@ describe('loadGitHubAppsFromEnvironment ', function () {
 
       GH_APP_IDENTIFIER: '42',
       GH_APP_SECRET_KEY: 'cheese',
-    }).load('__tmp_org_placeholder__').auth.appId;
+    }).get('__tmp_org_placeholder__').auth.appId;
     expect(actual).toEqual(42);
   });
 
--- a/src/config/loadGitHubAppsFromEnvironment.test.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.test.ts
@@ -12,9 +12,9 @@ describe('loadGitHubAppsFromEnvironment ', function () {
             auth: {
               appId: 42,
               privateKey: 'cheese',
+              installationId: undefined,
             },
             project: {
-              installationId: undefined,
               node_id: 'bread',
               product_area_field_id: 'wine',
               status_field_id: 'beer',
@@ -65,9 +65,9 @@ describe('loadGitHubAppsFromEnvironment ', function () {
             auth: {
               appId: 42,
               privateKey: 'cheese',
+              installationId: undefined,
             },
             project: {
-              installationId: undefined,
               node_id: 'PVT_kwDOABVQ184AOGW8',
               product_area_field_id: 'PVTSSF_lADOABVQ184AOGW8zgJEBno',
               status_field_id: 'PVTSSF_lADOABVQ184AOGW8zgI_7g0',
--- a/src/config/loadGitHubAppsFromEnvironment.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.ts
@@ -68,7 +68,7 @@ export function loadGitHubAppsFromEnvironment(env) {
 
   if (env.GH_APP_IDENTIFIER && env.GH_APP_SECRET_KEY) {
     // Collect config by (proleptic) envvar number. Once we have GH_APP_1_FOO
-    // this will make more sense. We'll collect shtuff in config and then
+    // this will make more sense. We'll collect stuff in config and then
     // instantiate a GitHubApp once all config has been collected for each
     // (once we've made a full pass through process.env).
 
--- a/src/config/loadGitHubAppsFromEnvironment.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.ts
@@ -64,7 +64,6 @@ export class GitHubAppsRegistry {
   constructor(configs) {
     this.apps = new Map<string, GitHubApp>();
     for (const config of configs.configs.values()) {
-      // 🏛️🍕🍕
       this.apps.set(config.org, new GitHubApp(config));
     }
   }
--- a/src/config/loadGitHubAppsFromEnvironment.ts
+++ b/src/config/loadGitHubAppsFromEnvironment.ts
@@ -42,12 +42,6 @@ export class GitHubAppsRegistry {
     }
   }
 
-  pop(org) {
-    const app = this.load(org);
-    this.apps.delete(org);
-    return app;
-  }
-
   load(org) {
     const app = this.apps.get(org);
     if (app === undefined) {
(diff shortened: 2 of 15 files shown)


src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,38 +1,74 @@
 import { AppAuthStrategyOptions } from '@/types';
 
+interface ProjectOptions {
+  node_id: string;
+  product_area_field_id: string;
+  status_field_id: string;
+  response_due_date_field_id: string;
+}
```
chadwhitacre, line 8: Probs should move this to `types`.

chadwhitacre: Done in 42b301376e880677f110ae04c35cf7d75ba91f21.

src/config/loadGitHubAppsFromEnvironment.test.ts
```diff
@@ -7,13 +7,14 @@ describe('loadGitHubAppsFromEnvironment ', function () {
         [
           '__tmp_org_placeholder__',
           {
+            num: 1,
             org: '__tmp_org_placeholder__',
             auth: {
               appId: 42,
               privateKey: 'cheese',
+              installationId: undefined,
```
chadwhitacre, line 15: Drive-by bug fix.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,36 +1,72 @@
...
+    this.num = obj.num;
+    this.org = obj.org;
+    this.auth = obj.auth;
+    this.project = obj.project;
   }
 }
 
-export class GitHubAppsRegistry {
-  apps: Map<string, GitHubApp>;
+class GitHubAppsConfigHelper {
```
chadwhitacre, line 20: I split out a separate class for the config pass, it felt too messy lumped in with the app registry.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,36 +1,72 @@
...
+          errors.push(`${group}.${key}.`);
+        }
+      });
+      if (errors.length) {
+        allErrors[org] = errors;
+      }
+    }
+    if (Object.keys(allErrors).length) {
+      throw new Error(`Config missing: ${JSON.stringify(allErrors)}`);
+    }
```
chadwhitacre, line 57: It'd be good to add some tests for validation failures.

chadwhitacre: Added one in #524, could pull up here if worth it.

chadwhitacre: Not worth it. Since we only parse config if the auth envvars are present (`IDENTIFIER`, `SECRET_KEY`), and we have defaults for the project ones, there's nothing for `validateAll` to do. I've removed it in 5bfc74c605dcb957a85e1792e6c7bcbcc6ebaf98, we can bring it back in #524.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -58,12 +63,32 @@ export class GitHubAppsRegistry {
...
 export function loadGitHubAppsFromEnvironment(env) {
-  const apps = new GitHubAppsRegistry();
+  const configs = new GitHubAppsConfigHelper();
+  let config;
 
   if (env.GH_APP_IDENTIFIER && env.GH_APP_SECRET_KEY) {
-    const org = '__tmp_org_placeholder__';
-    apps.register(new GitHubApp(org, env));
+    // Collect config by (proleptic) envvar number. Once we have GH_APP_1_FOO
+    // this will make more sense. We'll collect shtuff in config and then
```
hubertdeng123, line 71:
```suggestion
    // this will make more sense. We'll collect stuff in config and then
```

chadwhitacre: I was being silly. 😛

Merged (it's rewritten anyway in #524).

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,42 +1,45 @@
...
   constructor() {
-    this.apps = new Map<string, GitHubApp>();
+    this.configs = new Map<number, object>();
   }
 
-  register(app) {
-    this.apps.set(app.org, app);
+  forNumber(num: number): object | undefined {
+    if (!this.configs.has(num)) {
+      this.configs.set(num, { num: num });
```
hubertdeng123, line 29: I'm a bit confused about this, could you elaborate why this is setting an object here? Is it because the object returned is by reference?

hubertdeng123: maybe function name could be more descriptive?

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -1,42 +1,45 @@
...
-  pop(org) {
-    const app = this.load(org);
-    this.apps.delete(org);
-    return app;
+export class GitHubAppsRegistry {
+  apps: Map<string, GitHubApp>;
+
+  constructor(configs) {
+    this.apps = new Map<string, GitHubApp>();
+    for (const config of configs.configs.values()) {
```
hubertdeng123, line 40: the `configs.configs.values()` here was unclear to me what it was, configs here is more accurately `GitHubAppsRegistry`? Can this be renamed to be `githubAppsRegistry` or something similar?

src/config/loadGitHubAppsFromEnvironment.test.ts
azaslavsky: 🚗

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -58,12 +63,32 @@ export class GitHubAppsRegistry {
...
+    // (once we've made a full pass through process.env).
+
+    config = configs.forNumber(1);
+    config.org = '__tmp_org_placeholder__';
+    config.auth = {
+      appId: Number(env.GH_APP_IDENTIFIER),
+      privateKey: env.GH_APP_SECRET_KEY.replace(/\\n/g, '\n'),
+    };
+    config.project = {
+      node_id: env.ISSUES_PROJECT_NODE_ID || 'PVT_kwDOABVQ184AOGW8',
```
azaslavsky, line 82: Maybe we should store the sentital values (which I assume are for our default repo from before) in a single obj or set of consts somewhere, so it's clear what we are defaulting to?

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: This is a pretty normal pattern for populating a cache, no? We're filing configs away according to their number. (The number will eventually come from envvars like `GH_APP_1_FOO`, see the description on #482 and the implementation in #524. For now on this PR there's only one config (using the old/existing envvars) and it's hardwired to "1" below.) The `{ num: num }` is essentially a constructor for a new config object.

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: They're removed on the next PR (#524). I've been preserving this `|| 'PVT_foo'` pattern temporarily, this is the last place we have it.

src/config/loadGitHubAppsFromEnvironment.ts
hubertdeng123: I guess what is unclear to me is why `{ num: num }` is needed instead of just the empty object `{}`. If we have the config num to lookup, do we need that as part of the config later on?

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: No, it's `GitHubAppsConfigHelper`. Renamed to `configHelper` in 219157b307014f98e6715aa4de02615d656a84be (along with other changes :)).

hubertdeng123 (approved):

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: Ah. I store it on the config for convenience in debugging, it's not strictly necessary. Hopefully the refactor in 219157b307014f98e6715aa4de02615d656a84be makes it less distracting.

chadwhitacre: Like, I've found by experience that having the thing something is indexed by also stored on the thing is a good habit, because when you have the thing and want to know what it is stored under ... well! ;-)

PR #524, opened 2023-07-13
Implement multi-org config

chadwhitacre:
Part of #482, after #523.

This implements parsing of multiple apps from the environment. I may decide to replace `__tmp_org_placeholder__` on a separate PR first.

- [x] replace `__tmp_org_placeholder__` throughout
- [x] possibly expand testing
- [x] better error handling on config errors (envvars instead of path keys, better formatting)
- [x] address reviews
- [ ] put protection around places where we want to require GETSENTRY_ORG
- [x] ~clean up names (`GH_APPS` ⇒ `GH_ORGS`, envvars, etc.)~ _punting_

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -2,33 +2,25 @@
 # Follow the instructions in README.md as how to add your personal
 # tokens and secrets in this file
 
-# GitHub App ID
-GH_APP_IDENTIFIER=''
-# Github App Webhook Secret to verify requests come from GitHub
-GH_WEBHOOK_SECRET=''
-# GitHub App Secret Key
-# NOTE: this *must* be on a single line, otherwise it will break
-# reading the secret key.
-GH_APP_SECRET_KEY='-----BEGIN RSA PRIVATE KEY----------END RSA PRIVATE KEY-----'
-# Personal (ie, for your own GitHub account) user access token with the `admin:org`,
-# `admin:org_hook`, `admin: repo_hook`, `project`, `public_repo`, `repo:status`, and
-# `repo_deployment` permissions enabled.
+# GitHub
 GH_USER_TOKEN=''
+GH_WEBHOOK_SECRET=''
 
-# Slack signing secret to verify requests come from Slack
-SLACK_SIGNING_SECRET=
-# Slack Bot user access token -> "Bot User OAuth Token"
-SLACK_BOT_USER_ACCESS_TOKEN=
+# GitHub Apps
+GH_APP_1_ORG_SLUG=''
+GH_APP_1_IDENTIFIER=''
+GH_APP_1_SECRET_KEY='-----BEGIN RSA PRIVATE KEY----------END RSA PRIVATE KEY-----'
+GH_APP_1_ISSUES_PROJECT_NODE_ID=''
+GH_APP_1_PRODUCT_AREA_FIELD_ID=''
+GH_APP_1_STATUS_FIELD_ID=''
+GH_APP_1_RESPONSE_DUE_DATE_FIELD_ID=''
+
+# Slack
+SLACK_SIGNING_SECRET=''
+SLACK_BOT_USER_ACCESS_TOKEN=''
 
 # The name of the personal repository in your personal organization being used for development.
 PERSONAL_TEST_REPO='eng-pipes-dev'
-
-# Application-specific configuration
-# Config for the "GitHub Issues Someone Else Cares About" project board
-ISSUES_PROJECT_NODE_ID=''
-STATUS_FIELD_ID=''
-PRODUCT_AREA_FIELD_ID=''
-RESPONSE_DUE_DATE_FIELD_ID=''
 
 # Silence some GCP noise
 DRY_RUN=true
--- a/.env.test
+++ b/.env.test
@@ -1,9 +1,25 @@
 # GitHub
 FORCE_USER_TOKEN_GITHUB_CLIENT="false"
 GH_USER_TOKEN="ghp_BLAHBLAHBLAH"
-GH_APP_IDENTIFIER="1234"
 GH_WEBHOOK_SECRET="webhooksecret"
-GH_APP_SECRET_KEY="top \nsecret\n key"
+
+# some functionality is org-agnostic
+GH_APP_1_ORG_SLUG="Enterprise"
+GH_APP_1_IDENTIFIER="1234"
+GH_APP_1_SECRET_KEY="top \nsecret\n key"
+GH_APP_1_ISSUES_PROJECT_NODE_ID="PVT_blahblah"
+GH_APP_1_PRODUCT_AREA_FIELD_ID="PVTSSF_flooflah"
+GH_APP_1_STATUS_FIELD_ID="PVTSSF_zimzimzim"
+GH_APP_1_RESPONSE_DUE_DATE_FIELD_ID="PVTF_zamzamzam"
+
+# other functionality is hardwired for getsentry
+GH_APP_2_ORG_SLUG="getsentry"
+GH_APP_2_IDENTIFIER="7890"
+GH_APP_2_SECRET_KEY="super \nsecret\n key"
+GH_APP_2_ISSUES_PROJECT_NODE_ID="PVT_flahflah"
+GH_APP_2_PRODUCT_AREA_FIELD_ID="PVTSSF_moomah"
+GH_APP_2_STATUS_FIELD_ID="PVTSSF_mizmizmi"
+GH_APP_2_RESPONSE_DUE_DATE_FIELD_ID="PVTF_mazmaza"
 
 # Slack
 SLACK_SIGNING_SECRET="slacksigningsecret"
--- a/src/api/github/getClient.test.ts
+++ b/src/api/github/getClient.test.ts
@@ -23,10 +23,10 @@ describe('getClient(ClientType.User)', function () {
   });
 });
 
-describe("getClient(ClientType.App, 'getsentry')", function () {
+describe("getClient(ClientType.App, 'Enterprise')", function () {
   beforeAll(async function () {
     octokitClass.mockClear();
-    await getClient(ClientType.App, 'getsentry');
+    await getClient(ClientType.App, 'Enterprise');
   });
 
   it('is instantiated twice', async function () {
@@ -43,7 +43,7 @@ describe("getClient(ClientType.App, 'getsentry')", function () {
       auth: {
         appId: 1234,
         privateKey: 'top \nsecret\n key',
-        installationId: 'installation-getsentry',
+        installationId: 'installation-Enterprise',
       },
     });
   });
--- a/src/api/github/getClient.test.ts
+++ b/src/api/github/getClient.test.ts
@@ -1,5 +1,6 @@
 import { createAppAuth } from '@octokit/auth-app';
 
+import { GETSENTRY_ORG } from '@/config';
 import { ClientType } from '@api/github/clientType';
 import { getClient } from '@api/github/getClient';
 import { OctokitWithRetries as octokitClass } from '@api/github/octokitWithRetries';
@@ -44,6 +45,21 @@ describe("getClient(ClientType.App, 'Enterprise')", function () {
         appId: 1234,
         privateKey: 'top \nsecret\n key',
         installationId: 'installation-Enterprise',
+      },
+    });
+  });
+});
+
+describe('getClient(ClientType.App, GETSENTRY_ORG)', function () {
+  it('also knows about getsentry', async function () {
+    octokitClass.mockClear();
+    const actual = await getClient(ClientType.App, GETSENTRY_ORG);
+    expect(octokitClass).toHaveBeenLastCalledWith({
+      authStrategy: createAppAuth,
+      auth: {
+        appId: 7890,
+        privateKey: 'super \nsecret\n key',
+        installationId: 'installation-getsentry',
       },
     });
   });
--- a/src/brain/gocdSlackFeeds/deployFeed.test.ts
+++ b/src/brain/gocdSlackFeeds/deployFeed.test.ts
@@ -516,7 +516,7 @@ describe('DeployFeed', () => {
   });
 
   it('post message with commits in deploy link for getsentry', async () => {
-    const octokit = await getClient(ClientType.App, 'Enterprise');
+    const octokit = await getClient(ClientType.App, GETSENTRY_ORG);
     octokit.repos.getContent.mockImplementation((args) => {
       if (args.owner != 'getsentry') {
         throw new Error(`Unexpected getContent() owner: ${args.owner}`);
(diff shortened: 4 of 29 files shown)


chadwhitacre:
https://github.com/getsentry/eng-pipes/pull/460/files#r1263058169

.env.test
```diff
@@ -1,9 +1,25 @@
...
+GH_APP_1_RESPONSE_DUE_DATE_FIELD_ID="PVTF_zamzamzam"
+
+# other functionality is hardwired for getsentry
+GH_APP_2_ORG_SLUG="getsentry"
+GH_APP_2_IDENTIFIER="7890"
+GH_APP_2_SECRET_KEY="super \nsecret\n key"
+GH_APP_2_ISSUES_PROJECT_NODE_ID="PVT_flahflah"
+GH_APP_2_PRODUCT_AREA_FIELD_ID="PVTSSF_moomah"
+GH_APP_2_STATUS_FIELD_ID="PVTSSF_mizmizmi"
+GH_APP_2_RESPONSE_DUE_DATE_FIELD_ID="PVTF_mazmaza"
```
chadwhitacre, line 22: This is nice because it gives us multi-org testing, we're kind forced into it.

src/api/github/getClient.test.ts
```diff
@@ -23,10 +23,10 @@ describe('getClient(ClientType.User)', function () {
   });
 });
 
-describe("getClient(ClientType.App, 'getsentry')", function () {
+describe("getClient(ClientType.App, 'Enterprise')", function () {
```
chadwhitacre, line 26: I checked and yes GitHub org slugs are case sensitive, you can have uppercase.

src/brain/issueLabelHandler/followups.ts
```diff
@@ -77,8 +77,7 @@ export async function updateCommunityFollowups({
     return;
   }
 
-  const owner = payload.repository.owner.login;
-  const octokit = await getClient(ClientType.App, owner);
+  const octokit = await getClient(ClientType.App, app.org);
```
chadwhitacre, line 80: This is a nice deduplication, we already have to parse out the org in `loadAppFromPayload` so we can centralize that logic (see below) and reuse the result.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -53,41 +80,72 @@ export class GitHubAppsRegistry {
...
 
   loadFromPayload(payload) {
-    // Soon we aim to support multiple orgs!
-    const org = '__tmp_org_placeholder__'; // payload?.organization?.login;
-    if (!org) {
-      throw new Error(`Could not find an org in ${JSON.stringify(payload)}.`);
+    // Org slug is differently accessed in org-scoped APIs vs. repo-scoped APIs
+    let org: string;
+    if (payload.organization) {
+      org = payload.organization.login;
```
chadwhitacre, line 86: This has a net effect of doing away with some error handling we previously had, but I think it's okay. See https://github.com/getsentry/eng-pipes/pull/460/files#r1263058169.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -53,41 +80,72 @@ export class GitHubAppsRegistry {
...
+    } else if (payload.repository?.owner?.type === 'Organization') {
+      org = payload.repository.owner.login;
+    } else {
+      throw new Error(
+        `Could not find an org in '${JSON.stringify(
+          payload.organization
+        )}' or '${JSON.stringify(payload.repository?.owner)}'.`
```
chadwhitacre, line 93: Should probs test this.

src/webhooks/pubsub/index.ts
```diff
@@ -58,7 +58,7 @@ export const pubSubHandler = async (
   // call security warning.
   // https://codeql.github.com/codeql-query-help/javascript/js-unvalidated-dynamic-method-call/
   if (typeof func === 'function') {
-    app = GH_APPS.load('__tmp_org_placeholder__');
+    app = GH_APPS.load(GETSENTRY_ORG);
     octokit = await getClient(ClientType.App, GETSENTRY_ORG);
```
chadwhitacre, line 62: Derp. I need to loop over all orgs here.

src/webhooks/pubsub/index.ts
```diff
@@ -58,7 +58,7 @@ export const pubSubHandler = async (
   // call security warning.
   // https://codeql.github.com/codeql-query-help/javascript/js-unvalidated-dynamic-method-call/
   if (typeof func === 'function') {
-    app = GH_APPS.load('__tmp_org_placeholder__');
+    app = GH_APPS.load(GETSENTRY_ORG);
```
chadwhitacre, line 61: Derp. I need to loop over all orgs here.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -53,42 +80,70 @@ export class GitHubAppsRegistry {
...
+
+  for (const [envvar, value] of Object.entries(env)) {
+    // Find app configuration in env, grouping by number (GH_APP_1_FOO).
+
+    const m = envvar.match(/^GH_APP_(\d+)_([A-Z_]+)$/);
+    if (m === null) {
+      continue;
+    }
+
+    const n = parseInt(m[1], 10);
```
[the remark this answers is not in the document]
chadwhitacre: Well, so far Sentry has acquired two other companies in 15 years ... though it's accelerating! :o) I think we can live with this for a while and revisit when it becomes a pain point. Maybe we'll have a web UI by then and we can configure through that.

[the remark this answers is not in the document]
chadwhitacre: > Is the idea that if I have multiple apps, I'd have N `GH_APP_N_ORG_SLUG`, and so on for each other env var, for each app?

But yes, that's the idea. And the point about acquisitions is that we have one app per org, and we (roughly) have one org per brand ~= company acquired.

src/brain/gocdSlackFeeds/deployFeed.test.ts
```diff
@@ -516,7 +516,7 @@ describe('DeployFeed', () => {
   });
 
   it('post message with commits in deploy link for getsentry', async () => {
-    const octokit = await getClient(ClientType.App, GETSENTRY_ORG);
+    const octokit = await getClient(ClientType.App, 'Enterprise');
```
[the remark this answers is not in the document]
chadwhitacre: We are [mocking](https://github.com/getsentry/eng-pipes/blob/0ca5a8ad0b4e37adba26dae1025171d20b635bb1/src/api/github/__mocks__/octokitWithRetries.ts) the replies. 🐭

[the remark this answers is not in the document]
chadwhitacre: Oooh ... this one wants to be `GETSENTRY_ORG` though, good catch!

We end up testing both `getClient` cases pretty thoroughly since both are called a ton throughout the test suite.

src/config/loadGitHubAppsFromEnvironment.ts
```diff
@@ -84,31 +117,51 @@ export class GitHubApps {
...
+    const noop = (x) => x;
```
hubertdeng123, line 130: nit: might be more accurate to name this as `identity_fn`?

.env.test
hubertdeng123: good call

hubertdeng123 (approved):
Looks pretty good to me, pending some small comments that are made. Going to give the approval here since I'm out tomorrow

hubertdeng123 (changes requested):
Actually, I guess I can't really approve if the env vars aren't yet in `ci.yml` yet since it'll break everything. One thing there is that it would be nice if these env vars don't come from GH secrets in case we want to run GH graphql queries locally. They're not sensitive and it'd be nice to have those somewhere to use directly.
```
ISSUES_PROJECT_NODE_ID
PRODUCT_AREA_FIELD_ID
STATUS_FIELD_ID
RESPONSE_DUE_DATE_FIELD_ID
```

src/config/loadGitHubAppsFromEnvironment.ts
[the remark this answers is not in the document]
chadwhitacre: > we could just have `GH_APP_ORG_SLUGS` and then have the RHS be comma separated, then at runtime check to make sure each `GH_APP_*` field has the same number of entries

Also, I did consider this, but I feel that this is less scalable, in that it would be harder to visually inspect and modify envvars of this form. Imagine a comma-separated list of 20 apps, and you want to make a modification to number 13, but all 20 of them look like `PVT_af1728e2bf9b7133` or whatever. So dense, so error-prone. I think the present system scales better up to ~20 or so orgs/apps. By the time we get there I believe our situation will be different enough that we should evaluate available options then.

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: Eh, maybe. Since the function is called `mod` below as in `modify`, `noop` seems fine since it is in fact performing no modification operation. :)

.env.test
chadwhitacre: Done in d86f107e8517a7b06fafda6522dba1c4d375f63f, also updates `.env.example`.

.env.example
```diff
@@ -2,33 +2,25 @@
 # Follow the instructions in README.md as how to add your personal
 # tokens and secrets in this file
 
-# GitHub App ID
-GH_APP_IDENTIFIER=''
-# Github App Webhook Secret to verify requests come from GitHub
-GH_WEBHOOK_SECRET=''
-# GitHub App Secret Key
-# NOTE: this *must* be on a single line, otherwise it will break
-# reading the secret key.
```
chadwhitacre, line 11: We don't need docs in here since tell people above to look for docs in the README.

src/api/github/getClient.test.ts
chadwhitacre: Done in 78f1fbfe60f8e07c7157dea2f0df1a433ebdfb67.

src/config/loadGitHubAppsFromEnvironment.ts
chadwhitacre: Done in 2a616cd75149940558d9809673111a7886563c54. Writing tests did end up resulting in a couple improvements. 👍

src/webhooks/pubsub/index.ts
chadwhitacre: Done in 834c2b484a6d1d2f4fa32a4abd9791d834da19a5.

src/webhooks/pubsub/index.ts
```diff
@@ -45,30 +58,23 @@
...
+    reply.send();
   } else {
-    func = async () => {}; // no-op
-    code = 400;
+    reply.code(204);
+    reply.send(); // before real work to avoid blocking
+    const now = moment().utc();
+    for (const [org, app] of GH_APPS.apps) {
+      const octokit = await getClient(ClientType.App, org);
+      await func(app, octokit, now);
```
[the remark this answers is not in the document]
chadwhitacre: 😭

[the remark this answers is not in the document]
chadwhitacre: In their "good" examples ([one](https://github.com/github/codeql/blob/ab1f341aa69cf0bd0da88867d7e5a3569926cdaa/javascript/ql/src/Security/CWE-754/examples/UnvalidatedDynamicMethodCallGood.js), [two](https://github.com/github/codeql/blob/ab1f341aa69cf0bd0da88867d7e5a3569926cdaa/javascript/ql/src/Security/CWE-754/examples/UnvalidatedDynamicMethodCallGood2.js)) they test for `typeof func === 'function'`, I guess we invert this conditional.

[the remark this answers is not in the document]
chadwhitacre: Done in 580ddd6bb1013c4a903e887dc181f7450a8eb3db.

chadwhitacre:
> put protection around places where we want to require GETSENTRY_ORG

I'm thinking about this one. The risk is that we might process a payload from a non-`getsentry` org through an endpoint that is only intended for `getsentry`—i.e., the endpoint operates on `getsentry` regardless of what org we might send from. 🤔

chadwhitacre:
Here are the places where we call `getClient` with `GETSENTRY_ORG`. At the very least I need to audit these to make sure none of these operate on the org in the payload, they only touch `getsentry`.

```
$ ag --ignore \*.test.ts 'getClient\(ClientType\.App'
src/utils/db/getFailureMessages.ts
43:      const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/api/github/getRelevantCommit.ts
16:    const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/api/github/getChangedStack.ts
16:    const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/brain/ghaCancel/index.ts
27:  const octokit = await getClient(ClientType.App, owner);

src/brain/issueLabelHandler/triage.ts
63:  const octokit = await getClient(ClientType.App, app.org);
116:  const octokit = await getClient(ClientType.App, app.org);

src/brain/issueLabelHandler/route.ts
84:  const octokit = await getClient(ClientType.App, app.org);
139:  const octokit = await getClient(ClientType.App, app.org);

src/brain/issueLabelHandler/followups.ts
80:  const octokit = await getClient(ClientType.App, app.org);
144:  const octokit = await getClient(ClientType.App, app.org);

src/brain/typescript/getProgress.ts
21:  const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/brain/notifyOnGoCDStageEvent/index.ts
276:  const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/brain/projectsHandler/project.ts
67:  const octokit = await getClient(ClientType.App, app.org);

src/brain/requiredChecks/getAnnotations.ts
85:  const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/brain/pleaseDeployNotifier/actionViewUndeployedCommits.ts
71:  const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/brain/gocdSlackFeeds/deployFeed.ts
81:    const octokit = await getClient(ClientType.App, GETSENTRY_ORG);

src/webhooks/pubsub/index.ts
72:      const octokit = await getClient(ClientType.App, org);
```

chadwhitacre:
I guess the opposite could also happen, an Octokit created with something other than `getsentry` is used for `getsentry`. Does this generally result in a permissions failure? 🤔

```
      const octokit = await getClient(ClientType.App, 'foo');
      const { data } = await octokit.blah.blah({
        owner: 'bar',
        repo: 'baz',
      });
```

chadwhitacre:
Okay, I've audited all of the places where we call `getClient`. Findings:

1. GitHub's REST APIs seem to all take an org as an argument, whereas their GraphQL API seemingly does not (I guess the node IDs and whatnot are at least globally unique).
1. Where we instantiate with `GETSENTRY_ORG`, we always call with `GETSENTRY_ORG`. 👍
1. Where we instantiate with `app.org`, we sometimes call with `GETSENTRY_ORG` (stalebot), and other times it's not clear to me yet (everything using `githubEventHelpers`).
1. I'm struggling against the instinct to significantly refactor this so that app and especially the GitHub event helpers are properties on some sort of `GitHubOrg` object.

chadwhitacre:
Scrapping for parts.

PR #527, opened 2023-07-19
Add `Waiting for: Product Owner` across all issues where a community member comments

hubertdeng123:
see title

### DIFF
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -44,13 +44,6 @@ function isContractor(payload) {
   return payload.comment.author_association === 'COLLABORATOR';
 }
 
-function isNotWaitingForCommunity(payload) {
-  const { issue } = payload;
-  return !issue?.labels.some(
-    ({ name }) => name === WAITING_FOR_COMMUNITY_LABEL
-  );
-}
-
 // Markers of State
 
 export async function updateCommunityFollowups({
@@ -69,7 +62,6 @@ export async function updateCommunityFollowups({
     isNotInARepoWeCareAboutForFollowups,
     isNotFromAnExternalOrGTMUser,
     isContractor,
-    isNotWaitingForCommunity,
     isFromABot,
   ];
 
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -45,8 +45,7 @@ function isContractor(payload) {
 }
 
 function isPullRequest(payload) {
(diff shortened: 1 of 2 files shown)


hubertdeng123:
This should not include PR's. Maybe in the future, but there are some gaps right now that need to be addressed before that happens. Primarily, PR's do not have Product Areas, so slack notifications will break.

src/brain/issueLabelHandler/followups.ts
```diff
@@ -44,11 +44,9 @@ function isContractor(payload) {
   return payload.comment.author_association === 'COLLABORATOR';
 }
 
-function isNotWaitingForCommunity(payload) {
+function isPullRequest(payload) {
   const { issue } = payload;
-  return !issue?.labels.some(
-    ({ name }) => name === WAITING_FOR_COMMUNITY_LABEL
-  );
+  return issue.pull_request ? true : false;
```
azaslavsky, line 49:
```suggestion
  return !!issue.pull_request;
```

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -159,6 +168,7 @@ describe('issueLabelHandler', function () {
       {
         ...makePayload(repo, undefined, username),
         comment: { author_association: membership },
+        ...(isPR ? { issue: { pull_request: {} } } : {}),
```
azaslavsky, line 171: nit: I *think* I get what you're doing here ("only if`isPR === true` add the field `issue: { pull_request: {} }` to the object I am building"), but its pretty hard to read in practice.

azaslavsky (approved):

hubertdeng123:
This may be annoying when a product owner removes the `Waiting for: Product Owner` label and wants the item backlogged. Then multiple people comment on the issue and automation adds back `Waiting for: Product Owner`. 🤔

chadwhitacre (approved):

chadwhitacre:
> This may be annoying when a product owner removes the `Waiting for: Product Owner` label and wants the item backlogged. Then multiple people comment on the issue and automation adds back `Waiting for: Product Owner`. 🤔

Ftr we had a conversation about this and decided to proceed as-is for now. If there's an active conversation going on (like every few minutes back and forth) the product owner can leave "Waiting for" on there until the conversation dies down. What happens when a user comments and "Waiting for: Product Owner" is already applied? That's a no-op, correct?

hubertdeng123:
> What happens when a user comments and "Waiting for: Product Owner" is already applied? That's a no-op, correct?

Correct.

PR #535, opened 2023-07-21
Make GETSENTRY_ORG an org instead of a string

chadwhitacre:
Part of #482, after #534.

Where I'm going with this is to have both `slug` and `octokit` live on a single `GitHubOrg` object, so that when we make calls we can use both together in calls that look like this:

```ts
await org.api.issues.addLabels({                                                                          
  owner: org.slug,
  repo: repo,
  issue_number: issueNumber,
  labels: [WAITING_FOR_PRODUCT_OWNER_LABEL],
});
```

Or this:

```ts
const { data } = await GETSENTRY_ORG.api.repos.compareCommits({
  owner: GETSENTRY_ORG.slug,
  repo: GETSENTRY_REPO_SLUG,
  base: message.refId,
  head: headSha,
});
```

That won't make crossed wires impossible, but I think it reduces the risk to an acceptable level.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -171,6 +171,19 @@ export const GH_USER_TOKEN = process.env.GH_USER_TOKEN || '';
 /**
  * Load GitHubOrgs. We [want to] support processing events coming at us from
  * multiple GitHub orgs and this is how we [plan to] encapsulate it all.
+ *
+ * Some of the logic in eng-pipes is meant *only* for the `getsentry` org
+ * (things related to CI/CD, mostly), so we want to have a global reference to
+ * its GitHubOrg that we can import and use around the codebase. We're fine
+ * with failing somewhat opaquely at runtime if it's not configured (at least,
+ * that's the status quo). The default org below is intended to accomplish
+ * that. We're also fine with trusting that there are no webhook routes
+ * pointing at these parts of the codebase by which events from other GitHub
+ * orgs could leak into `getsentry`. Eep.
+ *
+ * Other logic (mostly related to automations for issues and discussions)
+ * operates on whatever org--possibly `getsentry`--we find in the payloads from
+ * GitHub. For those we use the GH_ORGS registry.
  */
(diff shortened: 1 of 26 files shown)


src/config/index.ts
```diff
@@ -168,10 +169,27 @@ export const SENTRY_REPOS: Set<string> = new Set([
...
 
 /**
- * Load GitHubOrgs. We support processing events coming at us from multiple
- * GitHub orgs and this is how we encapsulate it all.
+ * Load GitHubOrgs. We [want to] support processing events coming at us from
+ * multiple GitHub orgs and this is how we [plan to] encapsulate it all.
  */
 export const GH_ORGS: GitHubOrgs = loadGitHubOrgsFromEnvironment(process.env);
+export const GETSENTRY_ORG =
+  process.env.GH_APP_IDENTIFIER && process.env.GH_APP_SECRET_KEY
```
hubertdeng123, line 177: What do you think about this logic for creating the placeholder org to live in `loadGitHubOrgsFromEnvironment`? There seems to be an conditional case there for `process.env.GH_APP_IDENTIFIER && process.env.GH_APP_SECRET_KEY`, so we can use that so this can be shortened to 

`GETSENTRY_ORG = GH_ORGS.get(process.env.GETSENTRY_ORG || 'getsentry')`

chadwhitacre: But then wouldn't it return the stub object for _any_ org slug? The purpose here is to populate a stub object for only `getsentry`, since that is hardwired around the app. We're okay with the app failing moderately opaquely at runtime if `GETSENTRY_ORG` is not configured, because there shouldn't be any GitHub events reaching those parts of the app that aren't relevant outside `getsentry`. I can add this in a comment if helpful.

chadwhitacre: Done in ae3f5bcc8bf8c4ba6c6b424191fca82d02272fe8. Look better?

hubertdeng123: fair enough

hubertdeng123 (approved):

PR #536, opened 2023-07-21
Store Octokit instance at org.api

chadwhitacre:
Part of #482.

We want to be able to have the org slug and the Octokit bound to the org in the same place, to help avoid crossing wires. This puts the client on the org and starts to use it in the GitHub helpers.


chadwhitacre:
Oops. Accidentally pushed main to this, hence the merge. Redo in #537.

PR #537, opened 2023-07-22
Store Octokit instance at org.api

chadwhitacre:
Redo of #534, part of #482.

### DIFF
--- a/src/api/github/getClient.ts
+++ b/src/api/github/getClient.ts
@@ -1,13 +1,8 @@
-import { createAppAuth } from '@octokit/auth-app';
-
 import { GETSENTRY_ORG } from '@/config/index';
 import { GH_USER_TOKEN } from '@/config/index';
-import { AppAuthStrategyOptions } from '@/types';
 
 import { ClientType } from './clientType';
 import { OctokitWithRetries } from './octokitWithRetries';
-
-const _CLIENTS_BY_ORG = new Map();
 
 function _getUserClient() {
   return new OctokitWithRetries({
@@ -15,19 +10,7 @@ function _getUserClient() {
   });
 }
 
-function _getAppClient(auth: AppAuthStrategyOptions) {
-  return new OctokitWithRetries({
-    authStrategy: createAppAuth,
-    auth,
-  });
-}
-
-/**
- * Return an Octokit client.
- *
- * Only orgSlug is required, as we can assume the GH App is installed org-wide.
- */
-export async function getClient(type: ClientType, orgSlug?: string | null) {
+export function getClient(type: ClientType, orgSlug?: string | null) {
   if (process.env.FORCE_USER_TOKEN_GITHUB_CLIENT == 'true') {
     return _getUserClient();
   }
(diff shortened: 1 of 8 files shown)


chadwhitacre:
I'm stuck on mocking `octokitWithRetries` and `getClient.test.ts`.

chadwhitacre:
Note to self: update not in one place in the test suite but at all the `getClient` calls.

src/api/github/getClient.ts
```diff
@@ -43,31 +26,6 @@ export async function getClient(type: ClientType, orgSlug?: string | null) {
...
-      // > Additionally, the SDK will take care of regenerating an installation
-      // > access token for you so you don't need to worry about the one hour
-      // > expiration.
-      //
-      // https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#using-the-octokitjs-sdk-to-authenticate-as-an-app-installation
-
-      _CLIENTS_BY_ORG.set(orgSlug, client);
-    }
-    return client;
+    return GETSENTRY_ORG.api; // ☢️
```
chadwhitacre, line 29: We move all of the logic for instantiating the `Octokit` client over to `GitHubOrg`, but we return it here so that at this point we don't have to change anything in how `getClient` is called throughout the app.

src/api/github/org.ts
```diff
@@ -1,17 +1,48 @@
...
 
-  constructor(config: GitHubOrgConfig) {
+  constructor(userToken: string, config: GitHubOrgConfig) {
     this.slug = config.slug;
     this.appAuth = config.appAuth;
     this.project = config.project;
+
+    // Call bindAPI ASAP. We can't call it here because constructors can't be
+    // async.
+    this.api = new OctokitWithRetries({ auth: userToken });
```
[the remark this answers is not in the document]
chadwhitacre: Done in 5ec29160de3cd131bb0b892bbe7301cae5b9e1f0. I clean up async in the test suite as part of getting the error catching to work.

chadwhitacre:
Passing locally, not sure what's up, this error usually means something is drastically wrong like a syntax error or something. 🤔 =

hubertdeng123:
`Error: @jest/core tried to access pretty-format, but it isn't declared in its dependencies; this makes the require call ambiguous and unsound.`
😳

src/api/github/org.test.ts
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  it('does not try to get an org installation', function () {
+    expect(octokitClass.apps.getOrgInstallation).toHaveBeenCalledTimes(0);
+  });
+});
+
+describe('bindAPI', function () {
+  beforeAll(async function () {
+    const org = await new GitHubOrg('cheese please', {
```
hubertdeng123, line 37: does this need `await`?

hubertdeng123: maybe some weird behavior related async/await is causing the error in CI?

chadwhitacre:
Retry in 691ce4a51faee940342adc9951d5b318aa465c12.

src/api/github/org.test.ts
chadwhitacre: Simplified in 691ce4a51faee940342adc9951d5b318aa465c12.

chadwhitacre:
Okay so not locally but in CI yarn/jest is [choking](https://github.com/getsentry/eng-pipes/actions/runs/5650920224/job/15308104295#step:7:5) on [`throw` inside `constructor`](https://github.com/getsentry/eng-pipes/pull/537/commits/0d3576b205d9e580ac8a8bed1ad3f4dd0eed640b).

chadwhitacre:
Okay, I figured it out. The `throw 'string'` was masking the issue. `throw Error('string');` does better. We're instantiating `GitHubOrg` in `config/index.ts` and there's no `GH_USER_TOKEN` in the test environment. Working as expected. :)

This stuff gets refactored (with tests) in 0ecf5e29122105546eda4cdd5f418109ee027586 on #541. Let me see if it makes sense to pull some of that up.

chadwhitacre:
Should be ready in 4d4277852461142ea2cbd06c28170d9d1b12a101.

src/api/github/org.test.ts
```diff
@@ -0,0 +1,63 @@
...
+      appAuth: {
+        appId: 422,
+        privateKey: 'so private',
+      },
+    });
+    octokitClass.mockClear();
+    org.bindAPI();
+  });
+
+  it('is instantiated once again', async function () {
```
hubertdeng123, line 45: should these still have async functions? I remember you removing it before previously

src/api/github/org.test.ts
```diff
@@ -0,0 +1,63 @@
...
+
+import { GETSENTRY_ORG } from '@/config';
+import { OctokitWithRetries as octokitClass } from '@api/github/octokitWithRetries';
+
+import { GitHubOrg } from './org';
+
+describe('constructor', function () {
+  beforeAll(async function () {
+    octokitClass.mockClear();
+    await new GitHubOrg({
```
hubertdeng123, line 11: same here, is the await needed?

hubertdeng123 (approved):

src/api/github/org.test.ts
chadwhitacre: I don't think they're necessary but I don't think they hurt. I figured I would make minimal changes at this point, can revisit later.

PR #539, opened 2023-07-23
Move helpers to GitHubOrg

chadwhitacre:
Part of #482, after #538.

This moves the rest of the helper functions to `GitHubOrg` methods, for better encapsulation.


src/api/github/org.ts
```diff
@@ -57,4 +57,210 @@ export class GitHubOrg {
     }
     return response;
   }
+
+  async addIssueToGlobalIssuesProject(
```
hubertdeng123, line 61: 👏

hubertdeng123 (approved):

PR #541, opened 2023-07-23
Simplify getClient now that we can

chadwhitacre:
Part of #482, after #540.

This surfaced for me that `GH_USER_TOKEN` is a personal access token for the @getsentry-bot account.

Screenshot 2023-07-23 at 3 44 09 PM

To use this with another org the bot will need to be a member of that org and the PAT will need to be authorized. This bot has a dozen PATs and who knows what other integrations with the `getsentry` repo. Technically we could add it to `codecov`, and that might be what we do in the short term. In the long run it seems like a new account like `eng-pipes-lurker` or something would limit our security surface area and better communicate what's going on.


chadwhitacre:
Is it really the case that we can't use an app installed in an org to check membership? Do we _have_ to use a PAT for a true user account that is a member of the org?

Original implementation of membership checking:

- #100 
- #137
- #139

chadwhitacre:
Here's where we switched from app auth to user: #307.

(Also, P.S. here's where we dropped repo auth: #157.)

src/api/github/org.ts
```diff
@@ -8,34 +8,49 @@ import {
...
-  constructor(userToken: string, config: GitHubOrgConfig) {
+  constructor(config: GitHubOrgConfig) {
     this.slug = config.slug;
     this.appAuth = config.appAuth;
     this.project = config.project;
 
     // Call bindAPI ASAP. We can't call it here because constructors can't be
     // async. Note that in testing this ends up being mocked as if it were
-    // bound, even though (afaict) we generally don't call bindAPI in test.
-    this.api = new OctokitWithRetries({ auth: userToken });
```
chadwhitacre, line 27: If this actually worked it was a happy accident.

src/utils/getOssUserType.ts
```diff
@@ -28,7 +27,7 @@ export async function getOssUserType(
     return null;
   }
 
-  const org = owner.login;
+  const orgSlug = owner.login;
```
chadwhitacre, line 30: Updating this variable name to match (new) usage throughout the rest of the codebase.

src/config/index.ts
```diff
@@ -164,9 +164,16 @@ export const SENTRY_REPOS: Set<string> = new Set([
...
+ * As far as we can tell, it's not possible to check private org membership
+ * from an app installation. Therefore, we use a Personal Access Token for a
+ * bot account that is itself an org member.
+ *
+ * If you set FORCE... we will *always* use user auth instead of app
+ * installation auth, to make dev life easier.
  */
 export const GH_USER_TOKEN = process.env.GH_USER_TOKEN || '';
+export const FORCE_USER_TOKEN_GITHUB_CLIENT =
+  process.env.FORCE_USER_TOKEN_GITHUB_CLIENT == 'true';
```
hubertdeng123, line 176:
```suggestion
  process.env.FORCE_USER_TOKEN_GITHUB_CLIENT == 'true' || process.env.FORCE_USER_TOKEN_GITHUB_CLIENT == '1';
```

src/api/github/org.ts
```diff
@@ -8,12 +8,23 @@ import {
   GitHubOrgConfig,
 } from '@/types';
 
+import { FORCE_USER_TOKEN_GITHUB_CLIENT } from '../../config';
```
hubertdeng123, line 11:
```suggestion
import { FORCE_USER_TOKEN_GITHUB_CLIENT } from '@/config';
```

hubertdeng123 (commented):
> Is it really the case that we can't use an app installed in an org to check membership? Do we have to use a PAT for a true user account that is a member of the org?

Seems weird if we wouldn't be able to use the app installed in an org to check membership, provided the app has permissions to do so.

src/api/github/org.ts
```diff
@@ -23,22 +34,23 @@ export class GitHubOrg {
...
   async bindAPI() {
-    // Use an Octokit not bound to an org to make an Octokit bound to our org.
-    // The docs say it's safe for Octokit instances to be long-lived:
-    //
-    // > Additionally, the SDK will take care of regenerating an installation
-    // > access token for you so you don't need to worry about the one hour
-    // > expiration.
-    //
-    // https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/authenticating-as-a-github-app-installation#using-the-octokitjs-sdk-to-authenticate-as-an-app-installation
+    if (FORCE_USER_TOKEN_GITHUB_CLIENT) {
```
hubertdeng123, line 46: For my understanding, is this easier for dev because then we can just use user PAT instead of setting up test org?

chadwhitacre: Yes, see #473.

chadwhitacre: I was going to add a comment but there's already a comment. :)

chadwhitacre: Expanded the comment in 836c0050c20a4d9d9a3d800799df4697aa068ef8.

src/api/github/org.ts
chadwhitacre: Done in 836c0050c20a4d9d9a3d800799df4697aa068ef8.

src/config/index.ts
chadwhitacre: Done in 836c0050c20a4d9d9a3d800799df4697aa068ef8.

chadwhitacre:
> Seems weird if we wouldn't be able to use the app installed in an org to check membership, provided the app has permissions to do so.

Agreed. #307 is where we settled on the current approach. I dug a little deeper but decided to accept the status quo vs. fully re-researching it.

hubertdeng123 (approved):

PR #548, opened 2023-07-25
Configure for Codecov

chadwhitacre:
Primary PR for #482! 💃 After #545.

## Before Deployment

- [x] Populate `GH_APP_PRIVATE_KEY_FOR_CODECOV` to GH prod environment.
- [x] Set a new webhook secret for getsantry ([race condition](https://github.com/getsentry/eng-pipes/pull/551#issuecomment-1650817303), derp), store it in 1password.
- [x] Configure webhook and webhook secret in CoveCod ... leave inactive.
 
## After Deployment

- [ ] [Activate webhook](https://github.com/organizations/codecov/settings/apps/covecod)!
- [ ] [Subscribe to events](https://github.com/organizations/codecov/settings/apps/covecod/permissions)!
- [ ] Try it out in `codecov/feedback`!

PR #550, opened 2023-07-26
Allow for flexible webhook secrets

chadwhitacre:
Part of #482, see also #548.

We want to use the same webhook for both getsantry and covecod, but the original one is lost to us now. Having an alternative envvar will allow for seamless transition to a new webhook, which we can store in 1password for future org expansion.


hubertdeng123:
oof, yeah storing in 1password would be nice for future orgs too

hubertdeng123 (approved):

PR #554, opened 2023-07-26
creating new webhook for the options-automator

kneeyo1:
This is a work in progress associated with [this pr ](https://github.com/getsentry/sentry/pull/53610). The configoptions script runs as part of a kubernetes deployment and I would like to post the logs from that script to slack. 

Ideally, that linked script from sentry would send post requests to the webhook defined here, which then pushes it to slack.


src/webhooks/options-automator/options-automator.ts
```diff
@@ -0,0 +1,102 @@
...
+    await bolt.client.chat.postMessage({
+      blocks: successBlock,
+      channel: FEED_OPTIONS_AUTOMATOR_CHANNEL_ID,
+      unfurl_links: false,
+    });
+    return await bolt.client.chat.postMessage({
+      blocks: failedBlock,
+      channel: FEED_OPTIONS_AUTOMATOR_CHANNEL_ID,
+      unfurl_links: false,
+    });
```
kneeyo1, line 97: CI tests are yelling bc it says it needs text, but 

```
export interface ChatPostMessageArguments extends WebAPICallOptions, TokenOverridable {
    channel: string;
    text?: string;
    as_user?: boolean;
    attachments?: MessageAttachment[];
    blocks?: (KnownBlock | Block)[];
    icon_emoji?: string;
    icon_url?: string;
    metadata?: MessageMetadata;
    link_names?: boolean;
    mrkdwn?: boolean;
    parse?: 'full' | 'none';
    reply_broadcast?: boolean;
    thread_ts?: string;
    unfurl_links?: boolean;
    unfurl_media?: boolean;
    username?: string;
}
```
Shows that its not necessary, confused

owensk: Verify which version of the module you're getting w/ Yarn. It's not showing the `?` in the CI window so you may be tethered to an older version where that wasn't optional.

src/webhooks/options-automator/options-automator.ts
```diff
@@ -0,0 +1,102 @@
...
+  } catch (err) {
+    console.log(err);
```
hubertdeng123, line 99: It would be nice to send this to Sentry
```suggestion
    Sentry.captureException(err)
```

src/webhooks/options-automator/index.ts
```diff
@@ -0,0 +1 @@
+export { handler } from './options-automator';
```
hubertdeng123, line 1: guessing this is in progress?

src/webhooks/options-automator/options-automator.ts
```diff
@@ -0,0 +1,102 @@
...
+      message.set_options.map((option) =>
+        slackblocks.markdown(
+          `Set \`${option.option_name}\` with value \`${option.option_value}\``
+        )
+      )
+    ),
+  ];
+
+  try {
+    // @ts-ignore
```
hubertdeng123, line 87: Is this here bc CI tests are yelling?

src/blocks/slackBlocks.ts
```diff
@@ -13,9 +21,23 @@ export function markdown(text: string): MrkdwnElement {
...
+}
+
 export function section(block: MrkdwnElement): KnownBlock {
   return {
     type: 'section',
     text: block,
   };
 }
+
+export function SectionBlock(fields: MrkdwnElement[]): KnownBlock {
```
mattgauntseo-sentry, line 38: Would be good to add this to the above `section()` function.

`export function section(block: MrkdwnElement, fields: MrkdownElement[])`

src/webhooks/options-automator/options-automator.ts
```diff
@@ -0,0 +1,102 @@
...
+    await bolt.client.chat.postMessage({
+      blocks: successBlock,
+      channel: FEED_OPTIONS_AUTOMATOR_CHANNEL_ID,
+      unfurl_links: false,
+    });
+    return await bolt.client.chat.postMessage({
```
mattgauntseo-sentry, line 93: Nit: I'd remove the return here, it doesn't look like it's used for anything.

src/webhooks/options-automator/options-automator.ts
mattgauntseo-sentry: I faintly recall needing to set text but it being ignored and just the blocks were shown.

src/config/index.ts
```diff
@@ -24,6 +24,8 @@ export const GOCD_ORIGIN =
   process.env.GOCD_ORIGIN || 'https://deploy.getsentry.net';
 export const FEED_DEPLOY_CHANNEL_ID =
   process.env.FEED_DEPLOY_CHANNEL_ID || 'C051ED5GLN4';
+export const FEED_OPTIONS_AUTOMATOR_CHANNEL_ID =
+  process.env.FEED_OPTIONS_AUTOMATOR_CHANNEL_ID || 'C05JJ5JNZAB';
```
fpacifici, line 28: Where did you pick this channel id?
#proj-options-automator seems to be `C04URUC21C5`

src/types/index.ts
```diff
@@ -65,7 +65,7 @@ export type CheckRun = EmitterWebhookEvent<'check_run'>['payload']['check_run'];
 //
 // Need to do this as we can't convert a string literal union to an array of literals
 export const CHECK_RUN_PROPERTIES = ['id', 'head_sha', 'html_url'] as const;
-export type CheckRunProperty = typeof CHECK_RUN_PROPERTIES[number];
+export type CheckRunProperty = (typeof CHECK_RUN_PROPERTIES)[number];
```
fpacifici, line 68: Why this change ?

src/webhooks/options-automator/options-automator.ts
```diff
@@ -0,0 +1,137 @@
...
+    const successBlock: KnownBlock[] = [
+        slackblocks.header(slackblocks.plaintext('✅ Successfully Updated Options: ✅')),
+        ...(message.channel_updated_options.length > 0
+            ? [
+                slackblocks.divider(),
+                slackblocks.section(slackblocks.markdown('*Channel updated options:* ')),
+                slackblocks.SectionBlock(
+                message.channel_updated_options.map((option) =>
+                    slackblocks.markdown(`channel updated \`${option}\``)
+                )
```
fpacifici, line 29: Channel update is not a useful information for the users. Please remove them entirely.
We only care about successful and failed changes in the option value.

src/config/index.ts
kneeyo1: this was my test channel id while I was testing locally, will correct it before merging.

src/types/index.ts
kneeyo1: seems like an autoformatting change I didn't make this change

src/webhooks/options-automator/options-automator.ts
kneeyo1: yeah pretty much

src/blocks/slackBlocks.ts
```diff
@@ -13,9 +21,31 @@ export function markdown(text: string): MrkdwnElement {
...
+      type: 'section',
+      text: block,
+    };
+  } else if (fields) {
+    return {
+      type: 'section',
+      fields: fields,
+    };
+  } else {
+    throw new Error('Either block or fields must be provided.');
```
hubertdeng123, line 49: Feels like this error should be thrown before undefined is passed in for both block and fields here.

hubertdeng123 (commented):
lot of codecov warnings, perhaps this could use more coverage?

src/blocks/slackBlocks.ts
kneeyo1: since section is already being used by other functions rewriting it is probably out of scope, spoke with Hubert and trying to make either parameter optional just makes the code too messy and unreadable, so I'm going to keep it as two functions

hubertdeng123 (approved):

PR #556, opened 2023-07-26
Fix FORCE_USER_TOKEN_GITHUB_CLIENT

chadwhitacre:
Regressed in #482.

### DIFF
--- a/src/api/github/index.ts
+++ b/src/api/github/index.ts
@@ -1,8 +1,6 @@
 import { Webhooks } from '@octokit/webhooks';
 import * as Sentry from '@sentry/node';
 
-import { GH_USER_TOKEN } from '../../config';
-
 import { OctokitWithRetries } from './octokitWithRetries';
 
 const githubEvents = new Webhooks({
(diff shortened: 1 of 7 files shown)


src/api/github/index.ts
```diff
@@ -19,7 +17,7 @@ function defaultErrorHandler(error) {
 
 githubEvents.onError(defaultErrorHandler);
 
-function makeUserTokenClient(token: string = GH_USER_TOKEN) {
+function makeUserTokenClient(token: string) {
```
mattgauntseo-sentry, line 20: Should this function be removed an instead require folks to use the src/spi/github/makeUserTokenClient file instead?

mattgauntseo-sentry (approved):

src/api/github/index.ts
chadwhitacre: Oops! Good catch, I was going back and forth. Gone in bf33f8094762254cac446c5eec1c6e54ea737800.

chadwhitacre:
> /Users/matt/Projects/Code/eng-pipes/src/utils/getOssUserType.ts

🧐

PR #558, opened 2023-07-27
Upload sourcemap bundles

mattgauntseo-sentry:
I think this is right.

The switch from `../src/` to `/src/` was based on the note here: https://docs.sentry.io/platforms/javascript/sourcemaps/uploading/typescript/#1-generate-source-maps


chadwhitacre (approved):
Yes please!

PR #563, opened 2023-07-28
Add throttling to github client

mattgauntseo-sentry:
I'm seeing a lot of errors due to secondary rate limits in eng-pipes coming from gocd events. I've been running eng-pipes locally on the paths that would run into this issue and can't see any obvious problems that would cause this from GoCD so I can't see a behavior to change to address this, so instead this PR adds throttling to our API calls (waits for quota and retries or errors for secondary rate limits).

This follows the recommended pattern in the docs here: https://octokit.github.io/rest.js/v19

H/T to @IanWoodard for nudging this direction

NOTE: Renamed octokitWithRetries since it now does more and the pattern in their docs is "MyOctokit" -> "EngPipesOctokit"


src/api/github/engpipesOctokit.ts
chadwhitacre, line 1: Nit: `engPipesOctokit` if we're going to have `EngPipesOctokit` as the class name.

chadwhitacre (approved):
I'm +1 on the approach and I like the rename to `EngPipesOctokit`. One nit and I haven't thoroughly reviewed or tested so grain of 🧂.

PR #570, opened 2023-08-04
Remove `Waiting for: Product Owner` on staff or collaborator comments

hubertdeng123:
If a staff member or contractor comments on an issue that is `Waiting for: Product Owner`, remove the label.

### DIFF
--- a/src/api/github/org.ts
+++ b/src/api/github/org.ts
@@ -157,11 +157,7 @@ export class GitHubOrg {
     await this.sendGraphQuery(modifyProjectIssueFieldMutation, data);
   }
 
-  async clearProjectIssueField(
-    itemId: string,
-    projectFieldOption: string,
-    fieldId: string
-  ) {
+  async clearProjectIssueField(itemId: string, fieldId: string) {
     const modifyProjectIssueFieldMutation = `mutation {
       clearProjectV2ItemFieldValue(
         input: {
@@ -177,7 +173,6 @@ export class GitHubOrg {
     }`;
     const data = {
       itemId,
-      projectFieldOption,
       fieldId,
     };
     await this.sendGraphQuery(modifyProjectIssueFieldMutation, data);
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -86,15 +86,14 @@ export async function updateFollowupsOnComment({
         issue_number: issueNumber,
         name: WAITING_FOR_PRODUCT_OWNER_LABEL,
       });
+      const itemId: string = await org.addIssueToGlobalIssuesProject(
+        payload.issue.node_id,
+        repo,
+        issueNumber
+      );
+
+      await org.clearProjectIssueField(itemId, org.project.fieldIds.status);
     }
-
-    const itemId: string = await org.addIssueToGlobalIssuesProject(
-      payload.issue.node_id,
-      repo,
-      issueNumber
-    );
-
-    await org.clearProjectIssueField(itemId, org.project.fieldIds.status);
   } else {
     const isWaitingForCommunityLabelOnIssue = isLabelOnIssue(
       WAITING_FOR_COMMUNITY_LABEL
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -42,22 +42,20 @@ function isPullRequest(payload) {
 
 // Markers of State
 
-export async function updateCommunityFollowups({
+export async function updateFollowupsOnComment({
   id,
   payload,
   ...rest
 }: EmitterWebhookEvent<'issue_comment.created'>) {
   const tx = Sentry.startTransaction({
     op: 'brain',
-    name: 'issueLabelHandler.updateCommunityFollowups',
+    name: 'issueLabelHandler.updateFollowupsOnComment',
   });
 
   const org = GH_ORGS.getForPayload(payload);
 
   const reasonsToDoNothing = [
     isNotInARepoWeCareAboutForFollowups,
-    isNotFromAnExternalOrGTMUser,
-    isCommentFromCollaborator,
     isWaitingForSupport,
     isPullRequest,
     isFromABot,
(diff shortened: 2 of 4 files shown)


src/api/github/org.ts
```diff
@@ -157,11 +157,7 @@ export class GitHubOrg {
     await this.sendGraphQuery(modifyProjectIssueFieldMutation, data);
   }
 
-  async clearProjectIssueField(
-    itemId: string,
-    projectFieldOption: string,
```
hubertdeng123, line 162: `projectFieldOption` wasn't doing anything.

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -738,7 +760,7 @@ describe('issueLabelHandler', function () {
     it('should not modify labels when community member comments and issue is waiting for product owner', async function () {
       await setupIssue();
       await addLabel(WAITING_FOR_PRODUCT_OWNER_LABEL, 'routing-repo');
-      await addComment('routing-repo', 'Picard');
```
hubertdeng123, line 741: noticed this test case had the wrong user. `Picard` is an owner, not a community member

src/api/github/org.ts
```diff
@@ -157,11 +157,7 @@ export class GitHubOrg {
...
-    fieldId: string
-  ) {
+  async clearProjectIssueField(itemId: string, fieldId: string) {
```
azaslavsky, line 160: Codecov seems to think this never exercised by tests?

On further reflection: I guess you did a good thing by submitting a drive by fix here (removing an unused arg), and I'm punishing you by asking you to add a test too ("no good deed" etc etc)... So idk, I'm fine leaving this code as is, but the extra credit super duper brownie points version of this would be to break out this arg-removal into a separate PR where we also add a test to exercise this function. But I also know that's annoying and not on the critical path, so no pressure either way!

src/brain/issueLabelHandler/followups.ts
```diff
@@ -70,37 +68,65 @@ export async function updateCommunityFollowups({
...
-  const isWaitingForCommunityLabelOnIssue = payload.issue.labels?.find(
-    ({ name }) => name === WAITING_FOR_COMMUNITY_LABEL
-  )?.name;
+  const isLabelOnIssue = (labelName) =>
+    payload.issue.labels?.find(({ name }) => name === labelName)?.name;
 
-  if (isWaitingForCommunityLabelOnIssue) {
-    await org.api.issues.removeLabel({
+  if (
+    (await isNotFromAnExternalOrGTMUser(payload)) ||
```
azaslavsky, line 75: nit: its a bit confusing that one of these is `isNot...`, and the other `is...`. Took me a while to figure out what's what, would probably be a bit easier to read it we had `!isFromAnExternalOrGTMUser(payload) ||| isCommentFromCollaborator(payload)`.

Aside: what is a `GTMUser`?

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -691,6 +691,41 @@ describe('issueLabelHandler', function () {
       );
     });
 
+    it('should remove `Waiting for: Product Owner` label when staff member comments and issue is waiting for product owner', async function () {
```
azaslavsky, line 694: nit: "should remove `Waiting for: Product Owner` label when staff member comments and issue already has the `Waiting for: Product Owner` label"

Or am I misunderstanding this test, and `is waiting for product owner` refers to a condition other than already having that label?

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -691,6 +691,41 @@ describe('issueLabelHandler', function () {
...
+      await addLabel(WAITING_FOR_PRODUCT_OWNER_LABEL, 'routing-repo');
+      await addComment('routing-repo', 'Picard');
+      expect(org.api.issues._labels).toEqual(new Set(['Product Area: Test']));
+      expect(clearProjectIssueFieldSpy).toHaveBeenLastCalledWith(
+        'itemId',
+        org.project.fieldIds.status
+      );
+    });
+
+    it('should not remove `Waiting for: Community` label when staff member comments and issue is waiting for community', async function () {
```
azaslavsky, line 705: Ditto the wording change suggested above.

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -691,6 +691,41 @@ describe('issueLabelHandler', function () {
...
+      expect(org.api.issues._labels).toEqual(
+        new Set(['Product Area: Test', WAITING_FOR_COMMUNITY_LABEL])
+      );
+      expect(clearProjectIssueFieldSpy).not.toHaveBeenLastCalledWith(
+        'itemId',
+        org.project.fieldIds.status
+      );
+    });
+
+    it('should remove `Waiting for: Product Owner` label when collaborator comments and issue is waiting for product owner', async function () {
```
azaslavsky, line 718: Ditto.

src/brain/issueLabelHandler/index.test.ts
```diff
@@ -738,7 +773,7 @@ describe('issueLabelHandler', function () {
     it('should not modify labels when community member comments and issue is waiting for product owner', async function () {
       await setupIssue();
       await addLabel(WAITING_FOR_PRODUCT_OWNER_LABEL, 'routing-repo');
-      await addComment('routing-repo', 'Picard');
+      await addComment('routing-repo', 'Skywalker');
```
azaslavsky, line 776: How does this test still pass if we change the user adding the comment to the opposite of before?

Also, mixing Star Wars and Star Trek references is sacrilege!

hubertdeng123: Before, the functionality that existed was the same for when a user with either `OWNER` or `COLLABORATOR` permissions commented on an issue that was `Waiting for: Product Owner` (nothing should happen). Now, when an `OWNER` comments on an issue that is `Waiting for: Product Owner`, the label should be removed. That's why this test case was failing, since there was a typo

src/brain/issueLabelHandler/followups.ts
hubertdeng123: `GTMUser` is anyone from go to market! So that should cover sales and support

src/brain/issueLabelHandler/index.test.ts
hubertdeng123: you're understanding this correctly, I will change for more clarity

src/brain/issueLabelHandler/followups.ts
hubertdeng123: Going to not go with this change, as it is an imported function and is used in many places

src/api/github/org.ts
hubertdeng123: I would love to get those super duper brownie points, but this function itself doesn't return anything and writing a test for it doesn't seem to do much but increase the code coverage numbers. Typescript already complains if an arg is passed in that isn't used, but the only reason this slipped in was that the arg was being passed in for error reporting to Sentry in the case an exception was thrown. If I were to write a test, it would be asserting that the args being passed in are being used in the graph query, which doesn't always seem accurate given that eventually we may want to include more context for error reporting.

src/brain/issueLabelHandler/followups.ts
azaslavsky: Sgtm.

src/api/github/org.ts
azaslavsky: Makes sense!

azaslavsky (approved):

PR #575, opened 2023-08-08
Improve build cause filtering

mattgauntseo-sentry:
Noticed that the slack feed was intermittently showing the `Deploying <repo>@<sha>` in messages, this is because I grabbed the first build-cause rather than look for the git repo I wanted.


src/types/index.ts
```diff
@@ -171,6 +171,8 @@ type GoCDResultType = 'Passed' | 'Failed' | 'Cancelled' | 'Unknown';
 
 type GoCDStateType = 'Passed' | 'Failed' | 'Cancelled' | 'Building';
 
+export type GoCDBuildType = 'git' | 'pipeline';
```
IanWoodard, line 174: 👍

IanWoodard (approved):

PR #580, opened 2023-08-10
Sync product owners with `security-as-code`

hubertdeng123:
This writes to a yaml file in eng-pipes containing information relevant for automations regarding product owners.

Must go in before https://github.com/getsentry/security-as-code/pull/311

This is similar to the setup for sentry and sentry-docs. See here for more details:
https://github.com/getsentry/sentry/pull/48600

### DIFF
--- a/.github/workflows/react-to-product-owners-yml-changes.yml
+++ b/.github/workflows/react-to-product-owners-yml-changes.yml
@@ -1,0 +1,29 @@
+name: React to product-owners.yml changes
+on:
+  # This could be run manually, but the general expectation is that this fires
+  # from GHA in getsentry/security-as-code on changes there.
+
+  workflow_dispatch:
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    name: React to product-owners.yml changes
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: getsentry/action-setup-venv@v1.0.5
+        with:
+          python-version: 3.11.3
+
+      - name: Get an auth token
+        id: token
+        uses: getsentry/action-github-app-token@v2.0.0
+        with:
+          app_id: ${{ vars.SENTRY_INTERNAL_APP_ID }}
+          private_key: ${{ secrets.SENTRY_INTERNAL_APP_PRIVATE_KEY }}
+
+      - name: React to product-owners.yml changes
+        shell: bash
+        env:
+          GITHUB_TOKEN: ${{ steps.token.outputs.token }}
+        run: ./bin/react-to-product-owners-yml-changes.sh
--- a/bin/react-to-product-owners-yml-changes.py
+++ b/bin/react-to-product-owners-yml-changes.py
@@ -7,9 +7,6 @@ from os.path import dirname, join, realpath
 
 import yaml
 
-PREFIX = "Product Area: "
-
-
 def main():
     # Run from project root.
     os.chdir(realpath(join(dirname(sys.argv[0]), "..")))
--- a/bin/react-to-product-owners-yml-changes.py
+++ b/bin/react-to-product-owners-yml-changes.py
@@ -17,18 +17,27 @@ def main():
     product_owners_yml = yaml.safe_load(open(sys.argv[1]))
     product_area_to_team_map = {}
     team_to_slack_channel_map = {}
+    repo_to_team_map = {}
 
     for product_area, fields in product_owners_yml['by_area'].items():
         product_area_to_team_map[product_area] = fields['teams']
 
-    for team, fields in product_owners_yml['by_team']:
+    for team, fields in product_owners_yml['by_team'].items():
         # TODO: remove ternary here and require a slack channel for teams
         team_to_slack_channel_map[team] = fields['slack_channel'] if 'slack_channel' in fields else None
+
+    for repo, team in product_owners_yml['by_repo'].items():
+        repo_to_team_map[repo] = team
 
     fp = 'product-owners.yml'
 
     with open(fp, 'w') as file:
-        yaml.dump({ 'product_areas': product_area_to_team_map, 'teams': team_to_slack_channel_map}, file)
+        data = {
+            'product_areas': product_area_to_team_map,
+            'teams': team_to_slack_channel_map,
+            'repos': repo_to_team_map
+        }
+        yaml.dump(data, file)
 
     return 0
 
--- a/bin/react-to-product-owners-yml-changes.py
+++ b/bin/react-to-product-owners-yml-changes.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python3
+from __future__ import annotations
+
+import os
+import sys
+from os.path import dirname, join, realpath
+
+import yaml
+
+PREFIX = "Product Area: "
+
+
+def main():
+    # Run from project root.
+    os.chdir(realpath(join(dirname(sys.argv[0]), "..")))
+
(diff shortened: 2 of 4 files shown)


hubertdeng123:
tested locally and it looks good :)
```
product_areas:
  Alerts:
  - team-issues
  Crons:
  - team-crons
  Dashboards:
  - team-discover-and-dashboards
...
repos:
  arroyo: team-sns
  cdc: team-sns
  craft: team-ospo
  relay: team-ingest
  responses: team-ospo
  self-hosted: team-ospo
  sentry-android-gradle-plugin: team-mobile-sdks
  sentry-cocoa: team-mobile-sdks
  sentry-dart: team-mobile-sdks
  sentry-dart-plugin: team-mobile-sdks
...
teams:
  team-crons: null
  team-discover-and-dashboards: null
  team-docs: null
  team-enterprise: null
  team-ingest: null
  team-issues: null
  team-mobile-sdks: null
  team-ospo: C04B27EECUB

```

bin/react-to-product-owners-yml-changes.py
```diff
@@ -0,0 +1,46 @@
...
+import yaml
+
+PREFIX = "Product Area: "
+
+
+def main():
+    # Run from project root.
+    os.chdir(realpath(join(dirname(sys.argv[0]), "..")))
+
+    product_owners_yml = yaml.safe_load(open(sys.argv[1]))
```
azaslavsky, line 17: Bit confused here - what is this argument supposed to be? Or rather, I get that the it's the path to the product owners file, but what does that look like? Is it from another repo that we import as a dependency?

bin/react-to-product-owners-yml-changes.py
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env python3
```
azaslavsky, line 1: Would it be possible to add a test for this binary, even just one, just to give an idea of the expected usage? We could create a mock product-owners file and test against that. Otherwise this code interacts with some far off (closed source) repo in a way that's pretty difficult to deduce from the code alone.

bin/react-to-product-owners-yml-changes.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
+cd "$(dirname "$0")"
+cd ..
+gh repo clone https://github.com/getsentry/security-as-code /tmp/security-as-code
+sha=$(git -C /tmp/security-as-code rev-parse @)
+pip3 install pyyaml==6.0
+./bin/react-to-product-owners-yml-changes.py /tmp/security-as-code/rbac/lib/product-owners.yml
+branch="getsantry/update-product-areas-${sha:0:8}"
+message="Sync with product-owners.yml in security-as-code@${sha:0:8}"
```
azaslavsky, line 10: security-as-code is private - are we cool with potentially leaking SHAs like this? I'm not 100% sure myself, just curious. If we aren't, one alternative could just be to use the commit datetime instead?

bin/react-to-product-owners-yml-changes.sh
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env bash
+set -euo pipefail
```
azaslavsky, line 2: nit: `set -euox pipefail` will make the logs a bit nicer for us if things ever break.

bin/react-to-product-owners-yml-changes.sh
```diff
@@ -0,0 +1,19 @@
...
+cd "$(dirname "$0")"
+cd ..
+gh repo clone https://github.com/getsentry/security-as-code /tmp/security-as-code
```
azaslavsky, line 5: Do we need to `cd` into the right repo after making the tmp file? Seems like this works without it, curious as to why

.github/workflows/react-to-product-owners-yml-changes.yml
```diff
@@ -0,0 +1,29 @@
...
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    name: React to product-owners.yml changes
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: getsentry/action-setup-venv@v1.0.5
+        with:
+          python-version: 3.11.3
```
azaslavsky, line 16: Maybe stick with 3.8, since that's what we use elsewhere.

bin/react-to-product-owners-yml-changes.py
hubertdeng123: Yep, it's from security-as-code. That repo is cloned during the `cascade-product-owners-yml-changes.yml`
https://github.com/getsentry/security-as-code/blob/main/.github/workflows/cascade-product-owners-yml-changes.yml

hubertdeng123: The PR's you've been reviewing in security as a code are being read here

bin/react-to-product-owners-yml-changes.sh
hubertdeng123: nope, we don't care about anything in the `security-as-code` repo except the `product-owners.yml` file, which is passed in as an arg for the `react-to-product-owners-yml-changes.py` script.

.github/workflows/react-to-product-owners-yml-changes.yml
hubertdeng123: We actually do not:
https://github.com/getsentry/sentry/blob/b82e82f00ca79a92813b02c46c71dcd2b20c5fba/.github/workflows/react-to-product-owners-yml-changes.yml

Think this is fine for now

bin/react-to-product-owners-yml-changes.py
hubertdeng123: Looks like https://github.com/getsentry/security-as-code/pull/311 has already gone in. If your concern is expected usage, let me generate the expected file here to clear up the ambiguity. I don't think there's an easy way to test just this and don't think it's super helpful to test in CI, unless our team is making changes to the file.

hubertdeng123: added!

.github/workflows/react-to-product-owners-yml-changes.yml
azaslavsky: I guess my broader point is that both of these should be 3.8, since main sentry codebases are, and we should try and keep things consistent. But its a nit in any case, and a broader fix like that is out of the scope of this PR.

product-owners.yml
```diff
@@ -0,0 +1,131 @@
+product_areas:
```
azaslavsky, line 1: Do we want our real PA/team structure here? I was just imagining a small test file with like 2 PAs, 2 teams, 1 additiona_user, etc? +1'ed the PR in any case.

bin/react-to-product-owners-yml-changes.py
azaslavsky: Thanks! Though see my comment on product-owners.yml, maybe we should have mock that's clearly a test file, rather than the actual data?

azaslavsky (approved):
LGTM, barring some minor optional nits re: the test file.

product-owners.yml
hubertdeng123: Yes, we do. We'll use this later on for other automations and it's not sensitive data so it's safe.

bin/react-to-product-owners-yml-changes.py
hubertdeng123: We would want the actual data to be here, and it's not sensitive so it's fine.

PR #597, opened 2023-09-06
Start sending product area and teams info to bigquery

hubertdeng123:
first part of https://github.com/getsentry/team-ospo/issues/175

This starts sending data to BigQuery so that Looker can filter github event data by teams/product area.

I've talked with the data team and they have added the columns `teams` and `product_are` to super-big-data.open_source.github_events.

Screenshot 2023-09-06 at 1 13 42 PM


.gitignore
```diff
@@ -21,3 +21,4 @@ lib
 !.yarn/versions
 
 github-orgs.local.yml
+product-owners.local.yml
```
azaslavsky, line 24: Is this a file you use for local testing? Is there a script to generate it or something?

src/utils/getTeams.test.ts
```diff
@@ -0,0 +1,23 @@
+import { getTeams } from './getTeams';
+
+describe('getTeams', function () {
+  it('should return empty array if no teams were found', () => {
+    expect(getTeams('repo-does-not-exist', undefined, 'getsentry')).toEqual([]);
+  });
+
+  it('should return array with one team if repo does not have routing', () => {
+    expect(getTeams('test-ttt-simple', undefined, 'getsentry')).toEqual(['team-ospo']);
```
azaslavsky, line 9: 1. It's a bit difficult to figure out what's being tested when looking at the test based on name alone. I guess this is a repo with no routing, but I'm not 100% sure?

azaslavsky: Ah nm, it's in product-owners.yml. May be worthwhile to say somewhere around here that that's where you go to look for the data under test.

src/utils/getTeams.ts
```diff
@@ -0,0 +1,14 @@
+import { PRODUCT_OWNERS_YML, GH_ORGS } from "@/config";
+
+export function getTeams(repo: string, productArea: string | undefined, org: string): string[] {
```
azaslavsky, line 3: I'm far from a typescript expert, but isn't the idiomatic spelling of an optional parameter `productArea?: string`? Or will that behave differently in some subtle way from what you have here?

src/utils/metrics.ts
```diff
@@ -290,6 +305,10 @@ export async function insertOss(
     data.updated_at = comment.updated_at;
     data.target_id = issue.number;
     data.target_type = 'issue';
+    data.product_area = issue.labels?.find((label) =>
+      label.name.startsWith(PRODUCT_AREA_LABEL_PREFIX)
```
azaslavsky, line 309: nit: would be nice to add coverage for this.

src/utils/metrics.ts
```diff
@@ -271,6 +277,15 @@ export async function insertOss(
         data.target_id = label.id;
         data.target_name = label.name;
         data.target_type = 'label';
+        if (label.name.startsWith(PRODUCT_AREA_LABEL_PREFIX)) {
+          data.product_area = label.name;
+        }
+        else{
+          data.product_area = issue.labels?.find((label) =>
+            label.name.startsWith(PRODUCT_AREA_LABEL_PREFIX)
+          )?.name || null;
```
azaslavsky, line 286: What happens if we leave the `|| null` out? Is it now `undefined`, and is there a functional difference between those two?

src/utils/metrics.ts
```diff
@@ -271,6 +277,15 @@ export async function insertOss(
         data.target_id = label.id;
         data.target_name = label.name;
         data.target_type = 'label';
+        if (label.name.startsWith(PRODUCT_AREA_LABEL_PREFIX)) {
+          data.product_area = label.name;
+        }
+        else{
```
azaslavsky, line 283: nit: do we need to format? `else{` looks odd.

src/utils/metrics.ts
hubertdeng123: No functional difference, I just liked the idea of sticking to null for this case to keep things consistent, as that is the default value of the field.

src/utils/getTeams.ts
hubertdeng123: This is not very accurate as this field should never be undefined, and null is a type of string. Removing the `undefined` here.

src/utils/metrics.ts
hubertdeng123: done

src/utils/metrics.ts
hubertdeng123: done

azaslavsky (approved):

.gitignore
hubertdeng123: yep, just a local file and no there is no script to generate it. It's just a copy of the product-owners.yml file that is shipped with this repo

PR #599, opened 2023-09-07
Send APIs status through Sentry Bot

sentaur-athena:
Creating a hook for Sentaur bot that when called with "apis" it shows status of all our APIs.

Right now it's just reading a random from from Sentry because I don't have the data created in Sentry. yet.


sentaur-athena:
closing in favor of https://github.com/getsentry/eng-pipes/pull/600

PR #603, opened 2023-09-11
Better error handling for getTeams function

hubertdeng123:
There needs to be better error-handling here, so adding that in


src/utils/getTeams.test.ts
```diff
@@ -28,4 +30,28 @@ describe('getTeams', function () {
       'team-enterprise',
     ]);
   });
+
+  it('should return empty array if org is codecov', () => {
```
azaslavsky, line 34: Why does this and the test below have the same behavior? Is this just to verify that codecov is itself an undefined team?

src/utils/getTeams.ts
```diff
@@ -6,11 +8,23 @@ export function getTeams(
   org: string
 ): string[] {
   const orgObj = GH_ORGS.get(org);
+  // TODO: include codecov
+  if (org === 'codecov') {
+    return [];
```
azaslavsky, line 13: Seems like this line is not hit by any tests?

azaslavsky (approved):

src/utils/getTeams.ts
hubertdeng123: my test case appears to be wrong, fixing it now

src/utils/getTeams.test.ts
hubertdeng123: Yeah, if any event comes from the codecov org let's just return an empty array for teams temporarily until we roll this out to them

src/utils/getTeams.ts
hubertdeng123: fixed now.

PR #605, opened 2023-09-11
Switch Codecov to withRouting

chadwhitacre:
Decided on this with @eliatcodecov.


hubertdeng123:
For slack notifications to work, we need to have a stub product area in Codecov defined in https://github.com/getsentry/security-as-code/blob/main/rbac/lib/product-owners.yml

hubertdeng123:
Otherwise, we will have slack notifications sent to the `team-ospo` GitHub channel for issues that come in that are `Waiting for: Product Owner` without a product owner label.

PR #607, opened 2023-09-13
Lint codebase

hubertdeng123:
These files must've been committed with the `--no-verify` flag, so linting was skipped.

### DIFF
--- a/package.json
+++ b/package.json
@@ -108,8 +108,7 @@
   },
   "lint-staged": {
     "*.ts": [
-      "eslint --cache --fix",
-      "yarn test --findRelatedTests"
+      "eslint --cache --fix"
     ],
     "*.md": [
       "markdownlint"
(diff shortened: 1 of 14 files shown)


azaslavsky (commented):
Is it possible to put in a test or builder somewhere that validates that linting is correct? Maybe we can do it as part of `pre-commit run`? I wonder how other repos do it.

hubertdeng123:
> Is it possible to put in a test or builder somewhere that validates that linting is correct? Maybe we can do it as part of pre-commit run? I wonder how other repos do it.

It is currently a part of the pre-commit run in the `package.json`. See https://github.com/getsentry/eng-pipes/blob/main/package.json#L109. I think it was just manually skipped by people who were committing changes.

hubertdeng123:
I've also been guilty of this, maybe we can remove the `yarn test --findRelatedTests` line then? I feel like that's what is causing most people to skip the pre-commit check, and if it runs in the CI it doesn't seem extra useful to run pre-commit

azaslavsky (approved):

PR #616, opened 2023-09-20
Add a "No GoCD deploys in x hours" message to dev-infra feed.

mattgauntseo-sentry:
This should be low risk, it will message the team channel for now, but we can move to team-engineering if it seems useful and we've proven it works as intended.

### DIFF
--- a/src/brain/gocdNoDeploysAlert/index.ts
+++ b/src/brain/gocdNoDeploysAlert/index.ts
@@ -1,0 +1,84 @@
+import { KnownBlock } from '@slack/types';
+
+import { gocdevents } from '@/api/gocdevents';
+import { bolt } from '@/api/slack';
+import * as slackblocks from '@/blocks/slackBlocks';
+import {
+  FEED_DEV_INFRA_CHANNEL_ID,
+  GOCD_ORIGIN,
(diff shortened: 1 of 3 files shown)


src/brain/gocdNoDeploysAlert/index.ts
```diff
@@ -0,0 +1,84 @@
...
+    pipeline.name
+  );
+  if (!lastDeploy) {
+    // Do nothing until we have at least one deploy
+    return;
+  }
+
+  const date = new Date(lastDeploy.stage_last_transition_time);
+  if (Date.now() - DEPLOYS_FAILING_LIMIT_MS < date.getTime()) {
+    return;
```
IanWoodard, line 71: We should add a test for this.

IanWoodard (changes requested):

src/brain/gocdNoDeploysAlert/index.ts
mattgauntseo-sentry: I had a test that was meant for this but was slightly broken / misconfigured.

IanWoodard (approved):
LGTM

PR #619, opened 2023-09-21
Fixing migrate

IanWoodard:
Adding a more clear migrate:latest script


mattgauntseo-sentry (changes requested):
Can you update the migrate command on the README too

hubertdeng123 (changes requested):
I think this will break .github/workflows/migration.yml. That will also need to be changed, are we sure we want to `migrate:latest` everytime?

IanWoodard:
> Can you update the migrate command on the README too

Noticed and fixed

IanWoodard:
> I think this will break .github/workflows/migration.yml. That will also need to be changed, are we sure we want to `migrate:latest` everytime?

I realized this after changing, went for a hybrid approach

mattgauntseo-sentry (approved):

hubertdeng123 (approved):

PR #625, opened 2023-09-22
Use product-owners.yml as source of truth for slack triage notifications

hubertdeng123:
This removes logic from `/notify-for-triage` slack subscriptions to product areas, to establish a single source of truth for GitHub issue triage notifications that expands beyond just Product Areas. We will now rely on the information given in `product-owners.yml` from security-as-code instead of information stored in our postgres database. This allows us to include all sentry repos for slack triage notifications. 

This must go in after:
1. https://github.com/getsentry/security-as-code/pull/374
2. https://github.com/getsentry/eng-pipes/pull/626

Closes: https://github.com/getsentry/team-ospo/issues/177


src/brain/issueLabelHandler/followups.ts
```diff
@@ -241,12 +241,15 @@ export async function ensureOneWaitingForLabel({
         WAITING_FOR_PRODUCT_OWNER_LABEL,
         issue.labels,
         repo,
-        payload.organization?.login
+        org.slug
```
azaslavsky, line 244: Easier to read, nice!

src/utils/getTeams.test.ts
```diff
@@ -4,16 +4,16 @@ import { getTeams } from './getTeams';
 
 // Check tests/product-owners.yml for data used for tests here
 describe('getTeams', function () {
-  it('should return empty array if no teams were found', () => {
-    expect(getTeams('repo-does-not-exist', 'getsentry')).toEqual([]);
+  it('should return team-ospo if no teams were found', () => {
```
azaslavsky, line 7: Wonder if its not worth making a new team, `team-unknown` or similar, specifically for handling this case? Otherwise our channel will potentially be flooded, mixing things that are a: actually our team's tasks with b: things that are not our team's responsibility, but have fallen through to us due to configuration issues.

src/utils/getTeams.test.ts
```diff
@@ -4,16 +4,16 @@ import { getTeams } from './getTeams';
...
+    expect(getTeams('repo-does-not-exist', 'getsentry')).toEqual(['team-ospo']);
   });
 
   it('should return array with one team if repo does not have routing', () => {
-    expect(getTeams('test-ttt-simple', 'getsentry')).toEqual(['team-ospo']);
+    expect(getTeams('test-ttt-simple', 'getsentry')).toEqual(['team-issues']);
```
azaslavsky, line 12: Is the team selected at random?

hubertdeng123: If you look at `test/product-owners.yml` you'll see
```
test-ttt-simple: team-issues
```

I could comment that if it makes it more clear

azaslavsky: I get that, but I'm wondering how that maps to a real repo with no routing - how does it pick the team in that case?

hubertdeng123: Ah, that's done in security-as-code:
https://github.com/getsentry/security-as-code/blob/main/rbac/lib/product-owners.yml#L305

Then, it's synced to this repo when changes are made there

src/utils/getTeams.test.ts
hubertdeng123: Technically, we are in charge of the process of making sure issues are appropriately routed to the teams that are in charge of these issues. I was thinking the default behavior would be to assign those issues to us, not to close, but to make sure they get to the appropriate teams that should own the issue. 

That being said, it made sense to me to assign `team-ospo` here since that means that we should figure out the correct team to route an issue to.

azaslavsky (approved):

PR #634, opened 2023-09-27
Bump sentry sdk versions

hubertdeng123:
Think this should give us Query Insights?

https://docs.sentry.io/product/performance/query-insights/


mattgauntseo-sentry (approved):

src/brain/gocdSlackFeeds/deployFeed.ts
```diff
@@ -1,3 +1,5 @@
+import '@sentry/tracing';
```
hubertdeng123, line 1: Suggested by @lforst 
https://github.com/getsentry/sentry-javascript/issues/4731#issuecomment-1126157027

lforst: Hi, that comment is a bit outdated. We deprecated the Sentry tracing package (it's just no-ops now) so we could remove it entirely.

PR #636, opened 2023-09-27
Patch routing issue by product area

hubertdeng123:
https://sentry.sentry.io/issues/4509116827/?project=5246761&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=0&utc=true

Fixes this

### DIFF
--- a/src/utils/businessHours.test.ts
+++ b/src/utils/businessHours.test.ts
@@ -299,7 +299,7 @@ describe('businessHours tests', function () {
     it('should not calculate SLO violation if label is not untriaged', async function () {
       const result = await calculateSLOViolationTriage(
         'Status: Test',
-        [{ name: 'Test' }],
+        [{ name: 'Product Area: Test' }],
         'routing-repo',
         GETSENTRY_ORG.slug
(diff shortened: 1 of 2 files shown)


src/utils/businessHours.test.ts
```diff
@@ -329,7 +329,7 @@ describe('businessHours tests', function () {
     it('should calculate SLO violation if label is waiting for product owner', async function () {
       const result = await calculateSLOViolationTriage(
         WAITING_FOR_PRODUCT_OWNER_LABEL,
-        [{ name: 'Test' }],
```
azaslavsky, line 332: Can we have a test somewhere that fails when name: `Test` like it was before, to add the fact that the prefix is required to the spec?

azaslavsky (changes requested):

hubertdeng123:
I'm going to close this, as I'm not quite sure why slack triage notifications aren't sending to #feed-ospo. Going to revert https://github.com/getsentry/eng-pipes/pull/636 instead.

PR #639, opened 2023-09-27
Remove /notify-for-triage handlers and use product-owners.yml as source of truth

hubertdeng123:
Follow up to https://github.com/getsentry/eng-pipes/pull/625

Two main things were added:
1. Strips product area prefix from labels when finding product <-> team mapping
2. Logs whenever hourly github triage notifications are sent for debugging purposes


src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -248,162 +247,157 @@ export const constructSlackMessage = (
...
+    return bolt.client.chat
+      .postMessage({
+        channel: channelId,
+        text: '👋 Triage Reminder ⏰',
+        blocks: messageBlocks,
+      })
+      .then(async () => {
+        // Would be helpful to have some logs around sending slack triage notifications for debugging
+        // eslint-disable-next-line no-console
+        console.log(`Sent slack triage message to ${channelId}.`);
```
hubertdeng123, line 368: Yes, I know this is not ideal, but I tested this locally using the exact same payload as I should get in prod and it works. Not sure why it's not sending in prod and this will help to determine why

azaslavsky (approved):

src/webhooks/pubsub/slackNotifications.ts
azaslavsky: nit: I would recommend leaving a TODO pointed at a bug so we don't miss this.

PR #652, opened 2023-10-10
fix(apis): Read from api schema repo

sentaur-athena:
We decided that this file is better to live in sentry-api-schema repo. The file content is the same so the bot will behave the same.

Context for repo change: https://github.com/getsentry/sentry/pull/56698/

### DIFF
--- a/src/brain/apis/getOwnershipData.ts
+++ b/src/brain/apis/getOwnershipData.ts
@@ -1,7 +1,7 @@
 import { GETSENTRY_ORG } from '@/config';
 
 const OWNERSHIP_FILE_PATH = 'api_ownership_stats_dont_modify.json';
-const SENTRY_API_SCHEMA = 'sentry-api-schema'
+const SENTRY_API_SCHEMA = 'sentry-api-schema';
 
 export default async function getOwnershipData() {
   const resp = await GETSENTRY_ORG.api.rest.repos.getContent({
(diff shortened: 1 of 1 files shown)


src/brain/apis/getOwnershipData.ts
```diff
@@ -1,12 +1,12 @@
-import { GETSENTRY_ORG, SENTRY_REPO_SLUG } from '@/config';
+import { GETSENTRY_ORG } from '@/config';
 
-const OWNERSHIP_FILE_PATH =
-  'src/sentry/apidocs/api_ownership_stats_dont_modify.json';
+const OWNERSHIP_FILE_PATH = 'api_ownership_stats_dont_modify.json';
```
sentaur-athena, line 3: https://github.com/getsentry/sentry-api-schema/blob/main/api_ownership_stats_dont_modify.json

src/brain/apis/getOwnershipData.ts
```diff
@@ -1,12 +1,12 @@
...
+const SENTRY_API_SCHEMA = 'sentry-api-schema'
```
mattgauntseo-sentry, line 4:
```suggestion
const SENTRY_API_SCHEMA = 'sentry-api-schema';
```

Another option is to inline this. The only reason we have the SENTRY_REPO_SLUG constant is to make testing easier, but in this case, this is a constant and can be inlined in the getContent call.

mattgauntseo-sentry (approved):

PR #653, opened 2023-10-11
Add GH Issue Scores

hubertdeng123:
This will enable sending weekly GH engagement scores to slack by querying BigQuery.

Screenshot 2023-10-12 at 3 46 54 PM

A new google cron job will be set up in the ops repo here: https://github.com/getsentry/ops/pull/8194

Initially, this will send to the team-ospo channel, which we'll change to the eng management channel. A part of https://github.com/getsentry/team-ospo/issues/176


src/utils/scores.ts
```diff
@@ -0,0 +1,82 @@
+import { BigQuery } from '@google-cloud/bigquery';
+
+import { PROJECT } from '@/config';
+
+const bigqueryClient = new BigQuery({ projectId: PROJECT });
+
+export async function getIssueEventsForTeam(team) {
```
hubertdeng123, line 7: Don't see a good way to test this specifically 😔

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            type: 'plain_text',
+          },
+          type: 'header',
+        },
+        {
+          text: {
+            text: `\`\`\`┌──────────────────────────────────────────────────────────────┐
+| Team                              │ GitHub Responses on Time |
+├──────────────────────────────────────────────────────────────┤
+|team-ospo                          | 0/0 (100%)               |
```
chadwhitacre, line 34: Add space and remove redundant `team-` prefix?

```suggestion
| ospo                               | 0/0 (100%)               |
```

hubertdeng123: done!

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            emoji: true,
+            text: '🗓️ Weekly Team GitHub Engagement Scores 🗓️',
+            type: 'plain_text',
+          },
+          type: 'header',
+        },
+        {
+          text: {
+            text: `\`\`\`┌──────────────────────────────────────────────────────────┐
+| Team                          │ GitHub Responses on Time |
```
chadwhitacre, line 97:
```suggestion
| Team                          │ % on Time |
```

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        },
+      ])
+      .mockReturnValue([]);
+    await triggerSlackScores(null, null);
+    expect(postMessageSpy).toHaveBeenCalledWith({
+      blocks: [
+        {
+          text: {
+            emoji: true,
+            text: '🗓️ Weekly Team GitHub Engagement Scores 🗓️',
```
chadwhitacre, line 89:
```suggestion
            text: '🗓️ Weekly GitHub Response Times by Team 🗓️',
```

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            type: 'plain_text',
+          },
+          type: 'header',
+        },
+        {
+          text: {
+            text: `\`\`\`┌──────────────────────────────────────────────────────────┐
+| Team                          │ GitHub Responses on Time |
+├──────────────────────────────────────────────────────────┤
+| ospo                          | 1/1 (100%)               |
```
chadwhitacre, line 99:
```suggestion
| ospo                          | 100 (1/1)               |
```

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+| ingest                        | 0/0 (100%)               |
+| null                          | 0/0 (100%)               |
```
chadwhitacre, line 101: Push n=0 to bottom?

hubertdeng123: done

hubertdeng123:
hmmm lemme add padding on the percentage number

hubertdeng123:
better!
Screenshot 2023-10-12 at 3 16 53 PM

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        },
+        {
+          text: {
+            text: `\`\`\`┌────────────────────────────────────────────────┐
+| Team                          │ % on Time      |
+├────────────────────────────────────────────────┤
+| ospo                          | 100 (1/1)      |
+| issues                        | 50  (1/2)      |
+| enterprise                    | 0   (0/1)      |
+| ingest                        | 100 (0/0)      |
```
chadwhitacre, line 102:
```suggestion
| ingest                        |   - (0/0)      |
```

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+| null                          | 100 (0/0)      |
```
chadwhitacre, line 103: "null" is weird ... what does it mean? Would "other" or "no team" be clearer?

hubertdeng123: It's the name I'm using to indicate that the `slack_channel` and `offices` fields in product-owners.yml is null. I don't think that other or no team would be clearer for that.

src/webhooks/pubsub/slackScores.test.ts
hubertdeng123: good idea

src/webhooks/pubsub/slackScores.test.ts
chadwhitacre: Oh, hrm ... well I clearly didn't understand what it means! 😅

We can only know what team an issue is for if we have a `slack_channel` and `offices`?

hubertdeng123: Not quite, it's a team I've used for other test suites. It's something that's useful to test to ensure eng-pipes doesn't break when slack notifications are being sent out to a team without a `slack_channel` or `offices`, but it's not useful for this PR at all. It's loaded in to the test suite on initialization which is why it appeared here. Does that make sense?

It's defined [here](https://github.com/getsentry/eng-pipes/blob/main/test/product-owners.yml#L42)

I have attempted to try and reload `product-owners.yml` for each test case with no luck and eventually gave up, so it's shared between test files right now 😔

src/utils/scores.ts
azaslavsky: You mean mocking out bigquery? Could we wrap it in an interface that only has methods for getting the stuff we need from BigQuery, then mock that out?

azaslavsky: Another idea is just to create a spy that verifies that the exact SQL we expect gets sent, and provides mock results.

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+          text: {
+            emoji: true,
+            text: '🗓️ Weekly GitHub Response Times by Team 🗓️',
+            type: 'plain_text',
+          },
+          type: 'header',
+        },
+        {
+          text: {
+            text: `\`\`\`┌────────────────────────────────────────────────┐
```
azaslavsky, line 31: nit: I think we should have a leading newline in the output (maybe a trailing one too?), if for no other reason than that it makes tests easier to read ;)

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+| Team                          │ % on Time      |
+├────────────────────────────────────────────────┤
+| ospo                          | -   (0/0)      |
```
azaslavsky, line 34: nit: results that are equivalent in all other sorting metric (just % on time, right?) should be sorted alphabetically. This will make tests more deterministic, and the message itself easier to parse for users.

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -0,0 +1,114 @@
...
+            type: 'plain_text',
+          },
+          type: 'header',
+        },
+        {
+          text: {
+            text: `\`\`\`┌────────────────────────────────────────────────┐
+| Team                          │ % on Time      |
+├────────────────────────────────────────────────┤
+| ospo                          | 100 (1/1)      |
```
azaslavsky, line 99: What if there's another team with the same percentage, but a different number of issues (ex: team foobar has 2/2, and team bazqux has 2/4)? Proposal: we should order first by % complete descending, then by number of issues descending, then by alphabetical order ascending. So, we would have:

Frank | 100% (2/2)
Grace | 100% (1/1)
Alice | 50% (2/4)
Charlie | 50% (1/2)
Boris | 0% (0/0)
Daniella | 0% (0/0)

Wdyt?

src/webhooks/pubsub/slackScores.ts
```diff
@@ -0,0 +1,90 @@
...
+    {
+      type: 'header',
+      text: {
+        type: 'plain_text',
+        text: '🗓️ Weekly GitHub Response Times by Team 🗓️',
+        emoji: true,
+      },
+    },
+  ];
+  let scoreBoard =
```
azaslavsky, line 53: Should the border lines be scaled to the ..._WIDTH constants from above?

src/webhooks/pubsub/slackScores.ts
```diff
@@ -0,0 +1,90 @@
...
+) => {
+  const teamScores: teamScoreInfo[] = await Promise.all(
+    Object.keys(PRODUCT_OWNERS_INFO['teams']).map(async (team: string) => {
+      const issueTriageEvents = await getIssueEventsForTeam(team);
+      const triagedEvents = issueTriageEvents.filter(
+        (issue) => issue.is_triaged
+      );
+      const score =
+        issueTriageEvents.length === 0
+          ? -1
```
azaslavsky, line 31: Why is this -1, and not 0? Am I correct in assuming that it's to represent the different between "team had no events" vs "team had events, but did 0 of them". If so, I'd do an optional type instead here, that way the type system will help you ensure that all edge cases and combinations are covered. Otherwise, it's very easy write a bug where we forget to check the -1 case and end up with funky math (made especially dangerous by JS/TSs lack of a uint type!).

azaslavsky (changes requested):

src/webhooks/pubsub/slackScores.test.ts
chadwhitacre: I'm with @azaslavsky. In fact, I thought I said this already but maybe not. :o)

src/webhooks/pubsub/slackScores.test.ts
chadwhitacre: Ah! So `null` only shows up in test suites and not in prod?

hubertdeng123: yep, it only appears in test suites

chadwhitacre: Got it, thanks!

chadwhitacre:
Do we want a link to ... somewhere? So people can drill down and potentially take action? Maybe like this?

> [Looker Dashboard](https://sentryio.cloud.looker.com/dashboards/401) ❧ [Project Board](https://github.com/orgs/getsentry/projects/76)

src/webhooks/pubsub/slackScores.ts
hubertdeng123: It was -1 instead of 0 since teams with 0/0 issues should be sorted below those with 0/1. However, with the idea where we should sort by number of issues each team has received descending, that idea seems to be better here. I'll change this.

src/webhooks/pubsub/slackScores.test.ts
hubertdeng123: good idea

hubertdeng123: pushed!

src/utils/scores.ts
hubertdeng123: Ended up just creating a spy that verifies the exact sql

hubertdeng123:
> Do we want a link to ... somewhere? So people can drill down and potentially take action? Maybe like this?

I thought about this. At the moment there isn't a good view where people can see which responses are not triaged on time. Teams may have relatively bad `% on time` but nothing left to triage, so there isn't a direct correlation there. Although I guess it is helpful to get teams to still see which issues are still not triaged so they can address them. Guess there's no harm in adding

chadwhitacre:
> I thought about this.

Fair point. I guess the question is, what do we expect people to do with this report?

hubertdeng123:
> Fair point. I guess the question is, what do we expect people to do with this report?

Mostly to increase visibility and to inform managers how well their team is responding to issues. Teams already have a way of tackling issues that are overdue through slack triage notifications

src/utils/scores.test.ts
```diff
@@ -0,0 +1,101 @@
...
+      issue_id: '1',
+      repository: 'routing-repo',
+      product_area: 'Issues',
+      triaged_dt: { value: '2023-10-13T16:53:15.000Z' },
+      triage_by_dt: { value: '2023-10-12T21:52:14.223Z' },
+    },
+  ];
+});
+
+jest.mock('@google-cloud/bigquery', () => ({
```
azaslavsky, line 13: Just to be clear: this mock applies to all test cases, not just the ones in this file? In other words, is there anywhere in these tests that we're testing against a live BigQuery instance?

src/webhooks/pubsub/slackScores.ts
```diff
@@ -26,7 +26,10 @@ export const triggerSlackScores = async (
       const triagedEvents = issueTriageEvents.filter(
         (issue) => issue.is_triaged
       );
-      const score = triagedEvents.length / issueTriageEvents.length;
+      const score =
+        issueTriageEvents.length === 0
+          ? -1
```
azaslavsky: nit: Still think this is much clearer as an Optional type, rather than a special number you have to check against (ex: not obvious from the type that -1 is the only special number). But not going to block the PR on it, either way works.

azaslavsky (approved):
LGTM assuming we're not calling to BigQuery's live API over the network in any tests.

src/utils/scores.test.ts
hubertdeng123: None of the tests are against a live BigQuery instance

src/webhooks/pubsub/slackScores.ts
hubertdeng123: The reason I didn't change it to be an Optional type is because it introduces some more complex logic to the sorting function that is used later. Otherwise, there needs to be some sort of conditional in the sorting logic specifically for when the variable is undefined. While, if we have a special number 0 for the 0/0 case, the team will be placed at the bottom of the scoreboard automatically. I will write a comment here so that it's more clear.

PR #659, opened 2023-10-17
fix(scoreboard): Don't show issues that are not yet due on weekly scores

hubertdeng123:
Noticed there seemed to be a lot of events on our scoreboard, I think this is the reason why


azaslavsky (approved):
Had some questions about the logic here, but I talked it over with Hubert offline, and it all makes sense: each "event" is actually a pair of `Waiting For: Product Owner` being added and some triaging event (issue closed, `Waiting For: Community` applied, etc). So, for example, if a single issue was replied to late 5 times, that creates 5 event pairs, so it will be counted as `0/5` rather than `0/1` or `0/0`.

PR #661, opened 2023-10-17
Send discussion info to bigquery

hubertdeng123:
This requires webhooks to send first from https://github.com/organizations/getsentry/settings/apps/getsantry/permissions

### DIFF
--- a/src/utils/metrics.test.ts
+++ b/src/utils/metrics.test.ts
@@ -29,202 +29,277 @@ import { insertOss } from './metrics';
 
 describe('metrics tests', function () {
   describe('insertOss', function () {
-    const defaultPayload: Record<string, any> = {
-      action: 'labeled',
-      sender: {
-        login: 'username',
-        id: 'user_id',
-      },
-      repository: {
-        full_name: 'getsentry/routing-repo',
-        name: 'routing-repo',
-        owner: {
-          type: 'Organization',
+    describe('lableling events', function () {
+      const defaultPayload: Record<string, any> = {
+        action: 'labeled',
+        sender: {
+          login: 'username',
+          id: 'user_id',
         },
-      },
-      organization: {
-        login: 'getsentry',
-      },
-      issue: {
-        number: 1234,
-        created_at: null,
-        updated_at: null,
-        labels: [{ name: 'Product Area: Test' }],
-      },
-      label: {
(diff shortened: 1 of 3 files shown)


src/utils/metrics.ts
```diff
@@ -359,15 +369,26 @@ export async function insertOss(
...
+    data.created_at = comment.created_at;
+    data.updated_at = comment.updated_at;
+    data.target_id = comment.id;
+    data.target_name = discussion.title;
+    data.target_type = 'discussion';
   } else {
     // Unknown payload event, ignoring...
     return {};
   }
-  data.teams = getTeams(
```
hubertdeng123, line 366: I think it's wrong to have a team assigned to every GitHub event, especially if we're only tracking it for `issue` and `issue_comment` event types

src/utils/metrics.test.ts
```diff
@@ -29,202 +29,277 @@ import { insertOss } from './metrics';
...
-      sender: {
-        login: 'username',
-        id: 'user_id',
-      },
-      repository: {
-        full_name: 'getsentry/routing-repo',
-        name: 'routing-repo',
-        owner: {
-          type: 'Organization',
+    describe('lableling events', function () {
```
azaslavsky, line 32: nut: spelling

src/utils/metrics.test.ts
```diff
@@ -29,202 +29,277 @@ import { insertOss } from './metrics';
...
-      });
-      await getLabelsTable().insert({
-        label_name: 'Product Area: Other',
-        channel_id: 'CHNLIDRND1',
-        offices: ['sfo'],
+      let dateNowSpy;
+      beforeAll(async () => {
+        dateNowSpy = jest
+          .spyOn(Date, 'now')
+          .mockImplementation(() => 1487076708000);
```
azaslavsky, line 65: nit: may be a bit more readable if we use moment to generate this timestamp?

azaslavsky (approved):

src/utils/metrics.test.ts
hubertdeng123: good catch

src/utils/metrics.test.ts
hubertdeng123: done

PR #667, opened 2023-10-19
Fix duplicate slack scores

hubertdeng123:
We were triggering slack scores scheduled job twice, one for each org. I didn't catch this locally

### DIFF
--- a/src/webhooks/pubsub/slackScores.ts
+++ b/src/webhooks/pubsub/slackScores.ts
@@ -18,101 +18,100 @@ const NUM_ROW_SPACES = 3;
 const TEAM_PREFIX = 'team-';
 
 export const triggerSlackScores = async (
-  __org: GitHubOrg,
+  org: GitHubOrg,
   __now: moment.Moment
 ) => {
-  const teamScores: teamScoreInfo[] = await Promise.all(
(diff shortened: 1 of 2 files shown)


src/webhooks/pubsub/slackScores.ts
```diff
@@ -18,101 +18,100 @@ const NUM_ROW_SPACES = 3;
...
-          : triagedOnTimeEvents.length / issueTriageEvents.length;
-      return {
-        team: team.slice(TEAM_PREFIX.length),
-        score,
-        eventsTriagedOnTime: triagedOnTimeEvents.length,
-        numEvents: issueTriageEvents.length,
-      };
-    })
-  );
+  if (org.slug === 'getsentry') {
```
azaslavsky, line 24: Moving the code to the right like this feels unnecessary. Its best practice to return early on exit conditions like this, to avoid this sort of nesting:
```
if (org.slug != 'getsentry') {
  return
}

// ... rest of your logic...
```

Will also make the diff a lot smaller too!

hubertdeng123: done.

azaslavsky (approved):

PR #669, opened 2023-10-23
fix(github-metrics): Default teams to be empty array

hubertdeng123:
fixes https://sentry.sentry.io/issues/4517593692/?project=5246761&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=0&utc=true
😬


hubertdeng123:
I guess my local schema I was testing with was slightly different than in prod, since mine was nullable...

hubertdeng123:
I think a more permanent solution involves creating a dev BigQuery instance that is accessibly by everyone that is synced with the prod schema. Right now I'm using my own GCP project as a dev instance since I don't have access to the dev environment project that has been used in the past.

PR #673, opened 2023-10-25
Add discussion metrics weekly slack message

hubertdeng123:
This adds a weekly discussion metrics slack message. It'll look something like this:
Screenshot 2023-10-24 at 5 10 07 PM

It'll show the top 5 most active discussions and 5 most active Sentaurs


src/utils/scores.ts
```diff
@@ -81,3 +81,54 @@ export async function getIssueEventsForTeam(team) {
...
+      discussions.type = 'discussion_comment'
+      AND timestamp_diff(
+        CURRENT_TIMESTAMP(),
+        discussions.created_at,
+        day
+      ) <= 7
+      AND discussions.user_type != 'external'
+      AND discussions.user_type != 'bot'
+    GROUP BY discussions.username
+    ORDER BY num_comments DESC
```
azaslavsky, line 123: Out of scope for this PR, but I wonder if there is some standard way of running a local in-memory instance of SQLite to query against, or something similar? Otherwise we end up with a ton of untested code: we validate that the exact text of our SQL code is what we expect, but not what that SQL code actually does (ie, does it get the exact rows in the table we want it to?).

I could imagine using in-memory SQLite as our backend, since this looks like bog-standard SQL with no BigQuery specific extensions in it.

Just a thought - not a blocker by any means, and definitely not a thing to do in this PR, but something to think about, since we've had a few cases of queries not returning the data set we expect them to.

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -45,67 +67,67 @@ describe('slackScores Tests', function () {
 | null                          |   - (0/0)      |
 | ospo                          |   - (0/0)      |
```
azaslavsky, line 68: Does this mean we always report ospo, even when no metrics are returned? Or is that just how the test is configured?

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -160,13 +182,208 @@ describe('slackScores Tests', function () {
...
+        {
+          title: 'Discussion 3',
+          repository: 'test-ttt-simple',
+          discussion_number: '003',
+          num_comments: 1,
+        },
+      ];
+      const discussionCommenters = [
+        {
+          username: 'luke_skywalker',
```
azaslavsky, line 229: Missing a golden opportunity to have one of the commenters be chewbacca:

```
Chewbacca, 3 hours ago:
--------
Raaaaa-ra-ra awwww raaaaa!

- Sent from my iPhone 
```

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -160,13 +182,208 @@ describe('slackScores Tests', function () {
...
+          username: 'han_solo',
+          num_comments: 1,
+        },
+      ];
+      getDiscussionEventsSpy.mockReturnValue({
+        discussions,
+        discussionCommenters,
+      });
+      await sendDiscussionMetrics();
+      // Columns with links in them may seem a bit off, because the links won't actually appear in slack
```
azaslavsky, line 242: Is there any way to fix this? Or does it basically require running the code through a markdown parser?

azaslavsky (approved):

src/webhooks/pubsub/slackScores.test.ts
hubertdeng123: it's just how the test is configured, it uses the teams in `test/product-owners.yml`

src/webhooks/pubsub/slackScores.test.ts
hubertdeng123: you're right 🤦‍♂️

src/webhooks/pubsub/slackScores.test.ts
hubertdeng123: It will require us to run the code through a markdown parser, looks nasty in tests but it's well formed in slack!

src/utils/scores.ts
hubertdeng123: Yeah, that'd be nice to just to get some validation that it's valid sql. Adding to the backlog of things to do for eng-pipes:
https://github.com/getsentry/eng-pipes/issues/676

hubertdeng123: but yes, the dev environment setup for BigQuery leaves much to be desired....

I don't think most people who have worked in eng-pipes have even hooked up their env to BigQuery

PR #674, opened 2023-10-25
Add an alert for consecutive unsuccessful GoCD deploys

IanWoodard:
This should be a low risk change as it will only message #feed-dev-infra for now. This is similar to #616 but takes a slightly different approach which may prove to be more effective. In the future, if this proves useful, we could add it to other channels such as discuss-frontend and team-engineering.

### DIFF
--- a/src/brain/gocdConsecutiveUnsuccessfulAlert/index.test.ts
+++ b/src/brain/gocdConsecutiveUnsuccessfulAlert/index.test.ts
@@ -136,7 +136,7 @@ describe('gocdConsecutiveUnsuccessfulAlerts', function () {
       stage_state: 'Passed',
       stage_result: 'unknown',
       stage_create_time: new Date('2022-10-26T17:57:53.000Z'),
-      stage_last_transition_time: new Date(),
+      stage_last_transition_time: new Date('2022-10-26T17:57:53.000Z'),
       stage_jobs: '{}',
     });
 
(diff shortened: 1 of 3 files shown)


src/brain/gocdConsecutiveUnsuccessfulAlert/index.test.ts
```diff
@@ -0,0 +1,235 @@
...
+  });
+
+  afterAll(async function () {
+    await db.destroy();
+  });
+
+  beforeEach(async function () {
+    fastify = await buildServer(false);
+    await gocdConsecutiveUnsuccessfulAlert();
+    await db(DB_TABLE_STAGES).delete();
```
hubertdeng123, line 36: Is it expected to 
```
await db(DB_TABLE_STAGES).delete();
await db('slack_messages').delete();
```

on `beforeEach` and `afterEach`?

hubertdeng123 (approved):

src/brain/gocdConsecutiveUnsuccessfulAlert/index.test.ts
IanWoodard: I just followed the pattern used in https://github.com/getsentry/eng-pipes/blob/main/src/brain/gocdSlackFeeds/deployFeed.test.ts

PR #688, opened 2023-11-03
Send discussion metrics to #discuss-product

hubertdeng123:


src/webhooks/pubsub/slackScores.ts
```diff
@@ -169,7 +170,7 @@ export const sendGitHubEngagementMetrics = async (test: boolean = false) => {
   await Promise.all(slackNotifications);
 };
 
-export const sendDiscussionMetrics = async (__test: boolean = false) => {
+export const sendDiscussionMetrics = async (test: boolean = false) => {
```
azaslavsky, line 173: Does `test` mean "this is a test", or that we are doing a test run but on live code? I think maybe the naming should be clearer - maybe something like "ospo_internal"?

azaslavsky (approved):

src/webhooks/pubsub/slackScores.ts
hubertdeng123: sure, can to that `ospo_internal` is more clear

PR #697, opened 2023-11-08
Send options automator updates to Datadog

hubertdeng123:
Sends options automator updates as an event to Datadog through Events API.


src/webhooks/options-automator/options-automator.test.ts
```diff
@@ -26,246 +37,259 @@ describe('options-automator webhook', function () {
 
     expect(response.statusCode).toBe(200);
   });
-});
 
-describe('test message slack', function () {
-  afterEach(function () {
-    jest.clearAllMocks();
-  });
+  describe('messageSlack tests', function () {
```
hubertdeng123, line 41: these tests are unchanged, just moved stuff around in the file

src/webhooks/options-automator/options-automator.ts
```diff
@@ -1,21 +1,39 @@
...
 
+export async function sendOptionAutomatorUpdatesToDataDog(
+  message: OptionsAutomatorResponse
+) {
+  // Not sure how detailed we want the text message to be, but this should be fine initially
+  const params: v1.EventCreateRequest = {
+    title: 'Options Automator Update',
+    text: JSON.stringify(message),
+    dateHappened: moment.now(),
+  };
```
mwarkentin, line 33: Set an eventType? Seems like either `info` or `user_update` would make sense.. or maybe `error` or `success` depending on if the option update succeeded or not?

mwarkentin: Tags we could add if possible (not sure if we have the raw event data or not):

* `sentry_region` (us, de, disney, etc)
* `option_name` (`span-metrics-extraction-enabled`, `project-abuse-quota.attachment-limit`)
* `option_value`? Not sure if this makes sense
* `option_type` (`boolean`, `float`, etc.)

hubertdeng123: Do we want to send an options event for every option that is changed? Or rather, one event for when a group of options are changed?

Right now, we send a slack message whenever a group of options are changed

hubertdeng123: It seems reasonable to send an event when a group of options are changed, so that may make it difficult to have an `error` or `success` eventType. I'll use `info`. For tags, it seems reasonable to add the region name

hubertdeng123: It also seems reasonable to add the names of the options if that's useful

mwarkentin: Ideally they could be independent or 1 parent event with children events.

But if that's tricky to implement I'm good with 1 event with multiple options for now (although that will mean we can't tag it with any of the option values)

src/webhooks/options-automator/options-automator.ts
```diff
@@ -1,21 +1,41 @@
...
+export async function sendOptionAutomatorUpdatesToDataDog(
+  message: OptionsAutomatorResponse
+) {
+  // Not sure how detailed we want the text message to be, but this should be fine initially
+  const params: v1.EventCreateRequest = {
+    title: 'Options Automator Update',
+    text: JSON.stringify(message),
+    alertType: 'info',
+    dateHappened: moment.now(),
+    tags: [message.region],
```
mwarkentin, line 34: @hubertdeng123 Not sure how you do string interpolation in Typescript, but could you tag this like: `["sentry_region:${message.region}"]

hubertdeng123: yep, can do this

mwarkentin (approved):
Let's go!

src/webhooks/options-automator/options-automator.ts
kneeyo1: @mwarkentin this is what the current event emitted to dd looks like, I'm going to add the other base tags and stuff, but I think we could break down the events sent. I can add more logic here to see if there's any `errors` or `successes`, or just send every option as one event. 


image

PR #705, opened 2023-11-13
Datadog tags and events 

kneeyo1:
This PR does a couple things: 

- for each option changed, send an event to datadog 
- report as `success` or `error` depending on the option 
- adds the correct tags for datadog dashboard purposes and for tracking infra changes

### DIFF
--- a/src/webhooks/options-automator/options-automator.test.ts
+++ b/src/webhooks/options-automator/options-automator.test.ts
@@ -280,7 +280,7 @@ describe('options-automator webhook', function () {
   describe('sendOptionAutomatorUpdatesToDataDog tests', function () {
     it('should send the right payload', async function () {
       sendOptionAutomatorUpdatesToDataDog(testpayload, 1699563828);
-      expect(postMessageSpy).toHaveBeenCalledTimes(1);
+      expect(datadogApiInstanceSpy).toHaveBeenCalledTimes(1);
       expect(datadogApiInstanceSpy).toHaveBeenCalledWith({
         body: {
           dateHappened: 1699563828,
--- a/src/webhooks/options-automator/options-automator.test.ts
+++ b/src/webhooks/options-automator/options-automator.test.ts
@@ -280,13 +280,19 @@ describe('options-automator webhook', function () {
   describe('sendOptionAutomatorUpdatesToDataDog tests', function () {
     it('should send the right payload', async function () {
       sendOptionAutomatorUpdatesToDataDog(testpayload, 1699563828);
+      expect(postMessageSpy).toHaveBeenCalledTimes(1);
       expect(datadogApiInstanceSpy).toHaveBeenCalledWith({
         body: {
           dateHappened: 1699563828,
-          text: '{"region":"test_region","drifted_options":[{"option_name":"drifted_option_1","option_value":"value_1"},{"option_name":"drifted_option_2","option_value":"value_2"}],"updated_options":[{"option_name":"updated_option_1","db_value":"db_value_1","value":"new_value_1"}],"set_options":[{"option_name":"set_option_1","option_value":"set_value_1"},{"option_name":"set_option_2","option_value":"set_value_2"}],"unset_options":["unset_option_1","unset_option_2"],"not_writable_options":[{"option_name":"error_option_1","error_msg":"Error occurred for option 1"},{"option_name":"error_option_2","error_msg":"Error occurred for option 2"}],"unregistered_options":["unregisterd_option_1","unregisterd_option_2"],"invalid_type_options":[{"option_name":"invalid_type_option_1","got_type":"string","expected_type":"float"},{"option_name":"invalid_type_option_2","got_type":"float","expected_type":"int"}]}',
+          text: '"{"change":"region","option":"t"}"',
           title: 'Options Automator Update',
           alertType: 'info',
-          tags: ['sentry_region:test_region'],
+          tags: [
+            'sentry_region:st-test_region',
+            'source_tool:options-automator',
+            'source:options-automator',
+            'source_category:infra-tools',
+          ],
         },
       });
     });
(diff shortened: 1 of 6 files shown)


src/webhooks/options-automator/options-automator.ts
```diff
@@ -26,15 +26,49 @@ export async function sendOptionAutomatorUpdatesToDataDog(
...
+      const alertType = formatAlertType(optionType);
+
+      const params: v1.EventCreateRequest = {
+        title: 'Options Automator Update',
+        text: JSON.stringify(text),
+        alertType: alertType,
+        dateHappened: timestamp,
+        tags: [
+          region,
+          `source_tool:"options-automator"`,
```
tonyo, line 64: Do we actually need the `"` quotes there? Have some vague memories that it may be treated as part of the tag value.

kneeyo1: Don't think so actually. Taking them out

src/webhooks/options-automator/options-automator.test.ts
```diff
@@ -279,16 +279,42 @@ describe('options-automator webhook', function () {
 
   describe('sendOptionAutomatorUpdatesToDataDog tests', function () {
     it('should send the right payload', async function () {
-      sendOptionAutomatorUpdatesToDataDog(testpayload, 1699563828);
-      expect(datadogApiInstanceSpy).toHaveBeenCalledWith({
+      await sendOptionAutomatorUpdatesToDataDog(testparitalpayload, 1699563828);
```
hubertdeng123, line 282: nit: testPartialPayload

can we use camelcase here for our variables for test payloads?

src/webhooks/options-automator/options-automator.ts
```diff
@@ -26,15 +26,49 @@ export async function sendOptionAutomatorUpdatesToDataDog(
...
+        option: option,
+      };
+
+      const region = formatRegionTag(message.region);
+
+      const alertType = formatAlertType(optionType);
+
+      const params: v1.EventCreateRequest = {
+        title: 'Options Automator Update',
+        text: JSON.stringify(text),
```
hubertdeng123, line 59: I do think it would be helpful to use a more consise paragraph for `text`, I initially just used `JSON.stringify` because I didn't have enough time to write a custom message. It seems like since now an event is created for every option that is updated, it feels right to clean this up. 

Not blocking here, but just wanted to raise it up as potentially a TODO for the future

src/webhooks/options-automator/options-automator.ts
```diff
@@ -26,15 +26,49 @@ export async function sendOptionAutomatorUpdatesToDataDog(
...
 ) {
-  // Not sure how detailed we want the text message to be, but this should be fine initially
-  const params: v1.EventCreateRequest = {
-    title: 'Options Automator Update',
-    text: JSON.stringify(message),
-    alertType: 'info',
-    dateHappened: timestamp,
-    tags: [`sentry_region:${message.region}`],
+  const formatRegionTag = (region: string): string => {
+    if (region === 'us' || region === 'de') {
```
hubertdeng123, line 30: Can you separate this out into an const variable array of `SAAS_REGIONS` or something? Just so it makes it a bit easier to update in the future when we do add more regions

That would look something like
```
SAAS_REGIONS = ['us', 'de'];
...
if (SAAS_REGIONS.includes(region)) {...
```

src/webhooks/options-automator/options-automator.ts
```diff
@@ -26,15 +26,49 @@ export async function sendOptionAutomatorUpdatesToDataDog(
...
+      const alertType = formatAlertType(optionType);
+
+      const params: v1.EventCreateRequest = {
+        title: 'Options Automator Update',
+        text: JSON.stringify(text),
+        alertType: alertType,
+        dateHappened: timestamp,
+        tags: [
+          region,
+          `source_tool:options-automator`,
```
hubertdeng123, line 64: Since now we're sending an event per option changed, is it useful to have the option name as a tag?

src/webhooks/options-automator/options-automator.ts
```diff
@@ -26,15 +26,49 @@ export async function sendOptionAutomatorUpdatesToDataDog(
...
-  // Not sure how detailed we want the text message to be, but this should be fine initially
-  const params: v1.EventCreateRequest = {
-    title: 'Options Automator Update',
-    text: JSON.stringify(message),
-    alertType: 'info',
-    dateHappened: timestamp,
-    tags: [`sentry_region:${message.region}`],
+  const formatRegionTag = (region: string): string => {
+    if (region === 'us' || region === 'de') {
+      return `sentry_region:${region}`;
```
hubertdeng123, line 31: would be nice to have a test to ensure that `us` or `de` will be handled properly

hubertdeng123 (changes requested):

src/webhooks/options-automator/options-automator.ts
kneeyo1: can do something similar to what gets sent to slack?

kneeyo1: Yeah this can be a potential todo for the future. It would be a bit of refactoring.

hubertdeng123: can we create a GitHub Issue for this then? Just so we can track this and include a comment

`// TODO(getsentry/eng-pipes#144): Refactor Text Message`

src/webhooks/options-automator/options-automator.test.ts
```diff
@@ -234,7 +235,7 @@ describe('options-automator webhook', function () {
     });
     it('writes drift only', async function () {
       const postMessageSpy = jest.spyOn(bolt.client.chat, 'postMessage');
-      await messageSlack(testparitalpayload);
+      await messageSlack(testParitalPayload);
```
hubertdeng123, line 238: testParitalPayload -> testPartialPayload

hubertdeng123 (approved):

PR #708, opened 2023-11-14
adding the source to the payload, since the _admin tool will use this

kneeyo1:
This will help differentiate changes between the admin tool and the options-automator tool.


src/webhooks/options-automator/options-automator.ts
```diff
@@ -76,6 +76,10 @@ export async function sendOptionAutomatorUpdatesToDataDog(
 }
 
 export async function messageSlack(message: OptionsAutomatorResponse) {
+  if (message.source !== 'options-automator') {
```
hubertdeng123, line 79: Why is the payload being sent here, if it does not pertain to `options-automator`?

hubertdeng123: We want to send to DD, but not slack?

kneeyo1: if the way the option was changed wasn't by the options-auto (the other way is by the admin tool) then I dont want it to go to slack, since the channel is specifically for options-automator changes only

hubertdeng123 (approved):

hubertdeng123 (changes requested):
er actually, looks like a test is failing

kneeyo1:
> er actually, looks like a test is failing

Weird, they're passing locally for me. taking a look

hubertdeng123 (approved):

PR #712, opened 2023-11-28
Adding no deploys alert for s4s

IanWoodard:
Will hopefully help improve visibility around s4s failing to deploy (messages will only be sent in feed-dev-infra for now).


src/brain/gocdNoDeploysAlert/index.ts
```diff
@@ -18,6 +18,7 @@ import {
 import { getLastGetSentryGoCDDeploy } from '@/utils/db/getLatestDeploy';
 
 const PIPELINE_FILTER = [
+  'deploy-getsentry-backend-s4s',
```
hubertdeng123, line 21: Not a blocker, but is it worth putting this as a variable in `config/index.ts`?

hubertdeng123 (approved):

src/brain/gocdNoDeploysAlert/index.ts
IanWoodard: Will do this next time it is used

PR #723, opened 2023-12-07
feat(apis): schedule weekly slack message

sentaur-athena:
We want to send API stats to team-engineering every week for better visibility.

Testing using:
```
curl -X POST 'http://127.0.0.1:3000/webhooks/pubsub' -H "Content-Type: application/json" -d '{"message": {"data": "eyJuYW1lIjoic2xhY2stc2NvcmVzIn0="}}'
```

I can see the message in my test org:
Screenshot 2023-12-06 at 5 54 10 PM

### DIFF
--- a/src/brain/apis/index.ts
+++ b/src/brain/apis/index.ts
@@ -13,6 +13,45 @@ type SlackBlock = {
     text: string;
   };
 };
+
+export function getMessageBlocks(response, team = ''): Array<SlackBlock> {
+  const blocks: Array<SlackBlock> = [];
+    response.messages.forEach((message) => {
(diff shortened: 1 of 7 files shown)


src/brain/apis/index.ts
```diff
@@ -13,6 +13,45 @@ type SlackBlock = {
     text: string;
   };
 };
+
+export function getMessageBlocks(response, team = ''): Array<SlackBlock> {
```
sentaur-athena, line 17: Not new code. Just moving line 61+ to its own module so I can use it in SlackScores

PR #728, opened 2023-12-18
Uploading the original payload to sentry to better capture errors

kneeyo1:


src/webhooks/sentry-options/sentry-options.test.ts
```diff
@@ -42,6 +42,17 @@ describe('sentry-options webhook', function () {
       jest.clearAllMocks();
     });
 
+    it('handles errors and reports to Sentry', async function () {
+      jest
+        .spyOn(bolt.client.chat, 'postMessage')
+        .mockRejectedValue(new Error('Test Error'));
+      const sentryCaptureExceptionSpy = jest.spyOn(Sentry, 'captureException');
```
hubertdeng123, line 49: is Sentry imported? may need to do that if we're spying on it

src/webhooks/sentry-options/sentry-options.test.ts
```diff
@@ -42,6 +44,16 @@ describe('sentry-options webhook', function () {
       jest.clearAllMocks();
     });
 
+    it('handles errors and reports to Sentry', async function () {
+      const sentryCaptureExceptionSpy = jest
+        .spyOn(Sentry, 'postMessage')
```
hubertdeng123, line 49: this should probably still be `bolt.client.chat` instead of `Sentry`

src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -80,117 +79,117 @@ export async function messageSlack(message: SentryOptionsResponse) {
...
-    ),
-    ...(message.updated_options.length > 0
-      ? [
-          slackblocks.divider(),
-          slackblocks.section(slackblocks.markdown('*Updated options:* ')),
-          slackblocks.sectionBlock(
-            message.updated_options.map((option) =>
-              slackblocks.markdown(
-                `updated \`${option.option_name}\` with db value \`${option.db_value}\` to value \`${option.value}\``
+  try {
```
hubertdeng123, line 82: I think this try block was in the right place last time, feels like it should be done before sending the slack message, because that's really where a lot of unexpected behavior should occur. It should not happen in this part of the code

kneeyo1: I put it here because it'll catch a malformed message, and then report that message as a sentry error

hubertdeng123: I still don't think this should be necessary because there should be testing catching the malformed message before it is even sent out. However, if you do think there's a possibility of that happening, guess it is fine to keep here. I'm not super familiar with payloads being sent to this webhook.

hubertdeng123 (approved):

PR #729, opened 2023-12-20
Add issues to weekly github activity report

hubertdeng123:
Previously, we were only tracking discussions. Now, this adds issues to the report

### DIFF
--- a/src/utils/scores.test.ts
+++ b/src/utils/scores.test.ts
@@ -101,7 +101,7 @@ describe('score tests', () => {
 
   it('should send the right sql we expect for getDiscussionEvents', async () => {
     await getGitHubActivityMetrics();
-    const discussionsQuery = `
+    const discussionCommentsQuery = `
     SELECT
       discussions.target_name as title,
       discussions.repository as repository,
@@ -120,26 +120,26 @@ describe('score tests', () => {
     ORDER BY num_comments DESC
     ;`;
 
-    const githubCommentersQuery = `
+    const gitHubCommentersQuery = `
     SELECT
-      issues.username as username,
-      COUNT(issues.username) as num_comments,
+      comments.username as username,
+      COUNT(comments.username) as num_comments,
     FROM
-      \`open_source.github_events\` AS issues
+      \`open_source.github_events\` AS comments
     WHERE
-      (issues.type = 'discussion_comment' OR issues.type = 'issue_comment')
+      (comments.type = 'discussion_comment' OR comments.type = 'issue_comment')
       AND timestamp_diff(
         CURRENT_TIMESTAMP(),
-        issues.created_at,
+        comments.created_at,
         day
       ) <= 7
-      AND issues.user_type != 'external'
-      AND issues.user_type != 'bot'
-    GROUP BY issues.username
+      AND comments.user_type != 'external'
+      AND comments.user_type != 'bot'
+    GROUP BY comments.username
     ORDER BY num_comments DESC
     ;`;
 
-    const issuesQuery = `
+    const issueCommentsQuery = `
     SELECT
       issues.target_name as title,
       issues.repository as repository,
@@ -157,8 +157,8 @@ describe('score tests', () => {
     GROUP BY issues.target_name, issues.repository, issues.target_id
     ORDER BY num_comments DESC
     ;`;
-    expect(mockQuery).toHaveBeenCalledWith(discussionsQuery);
-    expect(mockQuery).toHaveBeenCalledWith(githubCommentersQuery);
-    expect(mockQuery).toHaveBeenCalledWith(issuesQuery);
+    expect(mockQuery).toHaveBeenCalledWith(discussionCommentsQuery);
+    expect(mockQuery).toHaveBeenCalledWith(gitHubCommentersQuery);
+    expect(mockQuery).toHaveBeenCalledWith(issueCommentsQuery);
   });
 });
--- a/src/utils/scores.test.ts
+++ b/src/utils/scores.test.ts
@@ -17,7 +17,7 @@ jest.mock('@google-cloud/bigquery', () => ({
     };
   },
 }));
-import { getDiscussionEvents, getIssueEventsForTeam } from './scores';
+import { getGitHubActivityMetrics, getIssueEventsForTeam } from './scores';
 
 describe('score tests', () => {
   afterEach(() => {
@@ -100,7 +100,7 @@ describe('score tests', () => {
   });
 
   it('should send the right sql we expect for getDiscussionEvents', async () => {
-    await getDiscussionEvents();
+    await getGitHubActivityMetrics();
     const discussionsQuery = `
     SELECT
       discussions.target_name as title,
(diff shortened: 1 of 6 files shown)


src/utils/scores.test.ts
```diff
@@ -120,25 +120,45 @@ describe('score tests', () => {
...
-    const discussionCommentersQuery = `
+    const githubCommentersQuery = `
     SELECT
-      discussions.username as username,
-      COUNT(discussions.username) as num_comments,
+      issues.username as username,
+      COUNT(issues.username) as num_comments,
     FROM
-      \`open_source.github_events\` AS discussions
+      \`open_source.github_events\` AS issues
```
azaslavsky, line 128: nit: `issues` -> `comments`? That seems to be what we're actually grabbing, if the `.type` values we filter for are anything to go by.

src/utils/scores.test.ts
```diff
@@ -120,25 +120,45 @@ describe('score tests', () => {
...
+        issues.created_at,
+        day
+      ) <= 7
+      AND issues.user_type != 'external'
+      AND issues.user_type != 'bot'
+    GROUP BY issues.username
+    ORDER BY num_comments DESC
+    ;`;
+
+    const issuesQuery = `
```
azaslavsky, line 142: nit: `issueCommentsQuery`. Took me a while that we're grabbing the comments, not the issues themselves.

Same for the tests below.

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -436,18 +472,27 @@ describe('slackScores tests', function () {
           num_comments: 1,
         },
       ];
-      getDiscussionEventsSpy.mockReturnValue({
+      const issues = [
+        {
+          title: 'Issue with `markdown`',
```
azaslavsky, line 477: Good thing to test for!

src/utils/scores.test.ts
```diff
@@ -120,25 +120,45 @@ describe('score tests', () => {
     ORDER BY num_comments DESC
     ;`;
 
-    const discussionCommentersQuery = `
+    const githubCommentersQuery = `
```
azaslavsky, line 123: Way out of scope for this PR, but I think we should consider exploring in-memory SQLite databases for testing, rather than matching these super long queries. Reasoning:

1. Most of the work this code is testing is actually in the query. So basically, this test boils down to "is the query spelled how it's spelled?". But we don't want to test the spelling of the query - we want to test the underlying semantics (given a table with these N rows, did it pick the subset S that we expect?).
2. My understanding is that BiqQuery SQL is a superset of SQLite's implementation of the SQL language. This is good: we can rely only on a small set of well-implemented and standardized features, rather than locking ourselves into the BigQuery ecosystem. BigQuery's extensions are probably very useful for efficient large scale (millions of rows) data processing, but at the scale we're operating at, using slightly less efficient SQL for greater portability and testability seems like a good trade.
3. I think tests written using an in-memory fake will be easier to read and verify for correctness. Reading through this query right now, I *think* it looks correct, but I'm basically just reading it verbatim and guessing at what it will do. If instead the test script was something like "create new in-memory DB, add 4 rows, run test, ensure only 2 rows showed up", it would be a lot easier to offer feedback ("you should add this kind of row", "it seems like your missing this kind of data", "we should add a row at the boundary of the `<= 7` condition", etc).

Anyway, what I propose is a rather large refactor, so I'm not sure it needs to be done now (or maybe ever), but just something to think about. Maybe a useful pattern to try for some simple query, either here or in the future, just to see how it feels?

src/webhooks/pubsub/slackScores.ts
```diff
@@ -198,17 +208,24 @@ export const sendGitHubEngagementMetrics = async (
...
-export const sendDiscussionMetrics = async (ospo_internal: boolean = false) => {
-  const { discussions, discussionCommenters } = await getDiscussionEvents();
-  if (discussions.length === 0 && discussionCommenters.length === 0) {
+export const sendGitHubActivityMetrics = async (
+  ospo_internal: boolean = false
+) => {
+  const { discussions, githubCommenters, issues } =
+    await getGitHubActivityMetrics();
+  if (
+    discussions.length === 0 &&
```
azaslavsky, line 217: nit: factor this compound condition into a separate variable, makes the code much more readable. Maybe 
```suggestion
    const noGitHubActivityFound = (!discussions && !githubCommenters && !issues)
    if (noGitHubActivityFound) {
```

src/webhooks/pubsub/slackScores.ts
```diff
@@ -198,17 +208,24 @@ export const sendGitHubEngagementMetrics = async (
   await Promise.all(slackNotifications);
 };
 
-export const sendDiscussionMetrics = async (ospo_internal: boolean = false) => {
-  const { discussions, discussionCommenters } = await getDiscussionEvents();
-  if (discussions.length === 0 && discussionCommenters.length === 0) {
+export const sendGitHubActivityMetrics = async (
+  ospo_internal: boolean = false
+) => {
+  const { discussions, githubCommenters, issues } =
```
azaslavsky, line 214: nit: is it `github` or `gitHub`? I'd argue for the latter, but in any case consistency makes the codebase easier to grep.

src/webhooks/pubsub/slackScores.ts
```diff
@@ -266,6 +283,38 @@ export const sendDiscussionMetrics = async (ospo_internal: boolean = false) => {
   scoreBoard += `└${'─'.repeat(
     DISCUSSION_COLUMN_WIDTH + NUM_COMMENTS_COLUMN_WIDTH + NUM_ROW_SPACES
   )}┘`;
+  firstColumnName = 'Most Active Issues this Week';
```
azaslavsky, line 286: Feels like the next 100 or so lines repeat the exact same functionality for each of discussions, issues, and gitHubCommenters. Could we factor our into a standalone helper function?

src/webhooks/pubsub/slackScores.ts
```diff
@@ -318,6 +367,6 @@ export const triggerSlackScores = async (
     return;
   }
   await sendGitHubEngagementMetrics();
```
azaslavsky, line 369: Unless I'm misunderstanding, you'd want these 3 calls to execute in parallel, rather than one after the other? So maybe:

```suggestion
  await Promise.all[sendGitHubEngagementMetrics(), sendDiscussionMetrics(), sendApisPublishStatus()];
```

src/utils/scores.test.ts
hubertdeng123: Yeah, you have brought this up before. I do think it would be nice to have but something that I don't think is worth it right now

src/webhooks/pubsub/slackScores.ts
hubertdeng123: changed to latter for consistency

src/webhooks/pubsub/slackScores.ts
hubertdeng123: done

src/webhooks/pubsub/slackScores.ts
hubertdeng123: I factored them out, it's still not super neat but seems to be a bit better than before.

src/utils/scores.test.ts
hubertdeng123: done

src/utils/scores.test.ts
hubertdeng123: done

src/webhooks/pubsub/slackScores.ts
```diff
@@ -247,99 +243,76 @@ export const sendGitHubActivityMetrics = async (
...
+    scoreBoard += `└${'─'.repeat(
+      DISCUSSION_COLUMN_WIDTH + NUM_COMMENTS_COLUMN_WIDTH + NUM_ROW_SPACES
+    )}┘`;
+    return scoreBoard;
+  };
+  const scoreBoard =
+    createTable(
+      discussions,
+      'Most Active Discussions this Week',
+      '# comments',
```
azaslavsky, line 301: nit: if this is the same param for all 3, can we get rid of it?

azaslavsky (approved):
Approved with one nit.

src/webhooks/pubsub/slackScores.ts
hubertdeng123: yeah, will remove

PR #731, opened 2023-12-21
Add additional repos for triage

hubertdeng123:
These repos are already defined in [security-as-code](https://github.com/getsentry/security-as-code/blob/main/rbac/lib/product-owners.yml#L394), so just adding them here

### DIFF
--- a/github-orgs.yml
+++ b/github-orgs.yml
@@ -45,6 +45,7 @@ getsentry:
       - 'sentry-symfony'
       - 'sentry-ruby'
       - 'sentry-cli'
+      - 'sentry-electron'
 
       # Mobile team T1
       # www.notion.so/346452f21e7947b4bf515d5f3a4d497d?v=cad7f04cf9064e7483ab426a26d3923a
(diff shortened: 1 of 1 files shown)


azaslavsky (approved):

github-orgs.yml
```diff
@@ -53,6 +54,9 @@ getsentry:
       - 'sentry-react-native'
       - 'sentry-unity'
       - 'sentry-dart'
+      - 'sentry-go'
+      - 'sentry-unreal'
+      - 'sentry-elixir'
```
cleptric, line 59: @hubertdeng123 Go, Elixir and Unreal belong to the web team and are T2 and T3. See https://www.notion.so/sentry/Hover-display-interactions-are-potentially-confusing-41980d94706746cf82fed0ae86da6e3e

Electron is also T2, not sure we should have added them here.

hubertdeng123: We've decided to expand triage beyond just T1 repos. See ticket here:
https://github.com/getsentry/team-ospo/issues/118

PR #735, opened 2023-12-22
For slack triage notifications, use business hours left to notify hourly

hubertdeng123:
It's a bit annoying when in the morning, a team gets surprised with a slack notification of an issue that is due in 24 minutes. This is because at the end of the day, we're not necessarily notifying teams of issues due the next morning since it is technically over 12 hours away. 

Also cleans up some unneeded async/await declarations

Closes this: https://github.com/getsentry/eng-pipes/issues/719

### DIFF
--- a/src/brain/issueLabelHandler/followups.ts
+++ b/src/brain/issueLabelHandler/followups.ts
@@ -234,12 +234,12 @@ export async function ensureOneWaitingForLabel({
   let timeToRespondBy;
   if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
     timeToRespondBy =
-      (await calculateSLOViolationTriage(
+      calculateSLOViolationTriage(
         WAITING_FOR_PRODUCT_OWNER_LABEL,
         issue.labels,
         repo,
         org.slug
-      )) || moment().toISOString();
+      ) || moment().toISOString();
   } else if (labelName === WAITING_FOR_SUPPORT_LABEL) {
     timeToRespondBy =
       (await calculateSLOViolationRoute(
--- a/src/utils/businessHours.test.ts
+++ b/src/utils/businessHours.test.ts
@@ -493,66 +493,61 @@ describe('businessHours tests', function () {
     it('should correctly calculate business hours left overnight', function () {
       const triageBy = '2023-12-22T18:00:00.000Z';
       const now = moment('2023-12-22T01:00:00.000Z');
+      const repo = 'test-ttt-simple';
+      const org = 'getsentry';
+      const productArea = 'Other';
       expect(
-        getBusinessHoursLeft(
-          triageBy,
-          now,
-          'test-ttt-simple',
-          'getsentry',
-          'Other'
-        )
+        getBusinessHoursLeft(triageBy, now, repo, org, productArea)
       ).toEqual(1);
     });
     it('should correctly calculate business hours over multiple days', function () {
       const triageBy = '2023-12-22T18:00:00.000Z';
       const now = moment('2023-12-21T01:00:00.000Z');
+      const repo = 'test-ttt-simple';
+      const org = 'getsentry';
+      const productArea = 'Other';
       expect(
-        getBusinessHoursLeft(
-          triageBy,
-          now,
-          'test-ttt-simple',
-          'getsentry',
-          'Other'
-        )
+        getBusinessHoursLeft(triageBy, now, repo, org, productArea)
       ).toEqual(9);
+    });
+    it('should correctly calculate business hours over holiday', function () {
+      const triageBy = '2024-01-02T17:00:00.000Z';
+      const now = moment('2023-12-23T00:00:00.000Z');
+      const repo = 'test-ttt-simple';
+      const org = 'getsentry';
+      const productArea = 'Other';
+      expect(
+        getBusinessHoursLeft(triageBy, now, repo, org, productArea)
+      ).toEqual(1);
     });
     it('should correctly account for weekends when calculating business hours', function () {
       const triageBy = '2023-01-17T18:00:00.000Z';
       const now = moment('2023-01-14T00:00:00.000Z');
+      const repo = 'test-ttt-simple';
+      const org = 'getsentry';
+      const productArea = 'Other';
       expect(
-        getBusinessHoursLeft(
-          triageBy,
-          now,
-          'test-ttt-simple',
-          'getsentry',
-          'Other'
-        )
+        getBusinessHoursLeft(triageBy, now, repo, org, productArea)
       ).toEqual(2);
     });
     it('should correctly calculate business hours for issues with non overlapping timezones', function () {
       const triageBy = '2023-12-22T18:00:00.000Z';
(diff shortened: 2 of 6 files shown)


src/brain/issueLabelHandler/followups.ts
```diff
@@ -234,12 +234,12 @@ export async function ensureOneWaitingForLabel({
   let timeToRespondBy;
   if (labelName === WAITING_FOR_PRODUCT_OWNER_LABEL) {
     timeToRespondBy =
-      (await calculateSLOViolationTriage(
```
azaslavsky, line 237: Kinda surprising that Typescript will let you await a function that's not a future. Woulda been nice if the compiler had caught this!

azaslavsky: Hmmm, upon further inspection I see that it is a future. How exactly does this work in practice then? Does calling the future without awaiting just trigger an async call? But then how does the variable get set?

src/utils/businessHours.test.ts
```diff
@@ -487,4 +488,72 @@ describe('businessHours tests', function () {
       );
     });
   });
+
+  describe('getBusinessHoursLeft', function () {
+    it('should correctly calculate business hours left overnight', function () {
+      const triageBy = '2023-12-22T18:00:00.000Z';
+      const now = moment('2023-12-22T01:00:00.000Z');
+      expect(
+        getBusinessHoursLeft(
```
azaslavsky, line 497: When a function signature is more than ~3 args, I think named parameters are the way to go. In particular, the last 3 arguments here are all the same type (strings), and as someone unfamiliar with the codebase, it's quite hard to figure out what corresponds to what just by reading the invocation. 

What if we did:
```
// At definition-site:
interface BusinessHoursLeft {
  triageBy: string,
  now: moment.Moment,
  repo: string,
  org: string,
  productArea: string
}
function getBusinessHoursLeft({ triageBy, now, repo, org, productArea }: BusinessHoursLeft {
  ...
}

// At call-site:
getBusinessHoursLeft({
          triageBy,
          now,
          repo: 'test-ttt-simple',
          org: 'getsentry',
          productArea: 'Other'
        })
```

Now I know what's what, and can catch errors that the type system can't (ex, swapped string args) more easily when scanning.

src/utils/businessHours.test.ts
```diff
@@ -487,4 +488,72 @@ describe('businessHours tests', function () {
...
+          triageBy,
+          now,
+          'test-ttt-simple',
+          'getsentry',
+          'Other'
+        )
+      ).toEqual(1);
+    });
+    it('should correctly calculate business hours over multiple days', function () {
```
azaslavsky, line 506: Maybe add a test for holidays as well?

src/utils/businessHours.test.ts
```diff
@@ -487,4 +488,72 @@ describe('businessHours tests', function () {
...
+          triageBy,
+          now,
+          'test-ttt-simple',
+          'getsentry',
+          'Other'
+        )
+      ).toEqual(9);
+    });
+    it('should correctly account for weekends when calculating business hours', function () {
+      const triageBy = '2023-01-17T18:00:00.000Z';
```
azaslavsky, line 520: Just curious what happens when we don't include milliseconds (or do include microseconds) in an ISO8601 timestamp. I know that Python's deserializer sometimes barfed on that, which caused trouble and test flakes in the past. Maybe JS's is smarter though.

src/brain/issueLabelHandler/followups.ts
azaslavsky: Nm - reading further still I see that you changed the definition too. Disregard :)

src/utils/businessHours.ts
```diff
@@ -177,3 +181,37 @@ export async function getNextAvailableBusinessHourWindow(
...
+  let momentIterator = moment(now.format());
+  const triageByMoment = moment(triageBy);
+  while (momentIterator.diff(triageBy, 'hours') < 0) {
+    const businessHourWindow = getNextAvailableBusinessHourWindow(
+      productArea,
+      momentIterator,
+      repo,
+      org
+    );
+    // Two cases to handle.
```
azaslavsky, line 202: Great comment!

src/brain/issueLabelHandler/followups.ts
hubertdeng123: It's calculation using the moment library to return a future timestamp. The functions themselves are not asynchronous anymore, so it's not needed. The reason why I had `await/async` there previously is because I had db operations somewhere baked in there, which are now removed.

src/webhooks/pubsub/slackNotifications.ts
```diff
@@ -152,12 +158,28 @@ export const constructSlackMessage = (
     const triageQueueIssues: SlackMessageUnorderedIssueItem[] = [];
     // Group issues into buckets based on time left until SLA
     let hasEnoughTimePassedSinceIssueCreation = false;
-    const addIssueToQueue = ({ url, number, title, triageBy, createdAt }) => {
+    const addIssueToQueue = ({
```
azaslavsky, line 161: Should we not modify [this](https://github.com/getsentry/eng-pipes/blob/f5d0058e4e449efa4e9cbaea4da9984510b03d28/src/webhooks/pubsub/slackNotifications.ts#L44) type definition to include `repo`, `org`, and `productArea` too?

hubertdeng123: that is already done above, typescript caught that

src/utils/businessHours.test.ts
hubertdeng123: The reason why I did not do this is because it seemed explicit enough what was going on where the function is being called:

```
const businessHoursLeft = getBusinessHoursLeft(
        triageBy,
        now,
        repo,
        org,
        productArea
      );
```

This ambiguity would only exist for testing. This is not a super widely used function, so I'd rather not declare an interface for it. I'll clear up what the strings are by giving them names before passing them in if that's helpful.

src/utils/businessHours.test.ts
hubertdeng123: this will be fine, since the moment library will parse the timestamp without milliseconds properly

src/utils/businessHours.test.ts
hubertdeng123: done

src/utils/businessHours.test.ts
azaslavsky: Sure, but that relies on the call-site naming its variables properly. If you use an interface, this is required, rather than optional, enforcing more readable code. The language even encourages this practice if you use an interface (you can just type the var name without stuttering), so it feels like a best practice.

In any case, I trust your judgment, just a small nit as a rather unfamiliar reader of the code.

hubertdeng123: There are other functions in the file that could use cleaning up too. Creating an issue to get those resolved in a later PR:
https://github.com/getsentry/eng-pipes/issues/737

azaslavsky (approved):

Issue #737, opened 2023-12-27
Name parameterize business hour functions


hubertdeng123:
For functions in `businessHours.ts` with more than 3 function params, we can name parameterize these

PR #738, opened 2023-12-29
Name parameterize business hour functions

hubertdeng123:
Closes https://github.com/getsentry/eng-pipes/issues/737

### DIFF
--- a/src/utils/businessHours.ts
+++ b/src/utils/businessHours.ts
@@ -25,13 +25,21 @@ interface BusinessHourWindow {
   end: moment.Moment;
 }
 
-export function calculateTimeToRespondBy(
-  numDays: number,
-  productArea: string,
-  repo: string,
-  org: string,
-  testTimestamp?: string
-): string {
+interface CalculateTimetoRespondByParams {
+  numDays: number;
+  productArea: string;
+  repo: string;
+  org: string;
+  testTimestamp?: string;
+}
+
+export function calculateTimeToRespondBy({
+  numDays,
+  productArea,
+  repo,
+  org,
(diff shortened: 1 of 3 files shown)


src/utils/businessHours.ts
```diff
@@ -53,29 +61,34 @@ export function calculateTimeToRespondBy(
...
   // calculate time to triage for issues that are rerouted
   else if (
-    target_name.startsWith(PRODUCT_AREA_LABEL_PREFIX) &&
+    targetName.startsWith(PRODUCT_AREA_LABEL_PREFIX) &&
     labels?.some((label) => label.name === WAITING_FOR_PRODUCT_OWNER_LABEL)
   ) {
-    return calculateTimeToRespondBy(
-      MAX_TRIAGE_DAYS,
-      target_name.slice(PRODUCT_AREA_LABEL_PREFIX.length),
+    return calculateTimeToRespondBy({
```
azaslavsky, line 86: Surprising that this line is not hit by tests? Maybe the instrumentation is being weird?

src/utils/businessHours.ts
```diff
@@ -25,13 +25,21 @@ interface BusinessHourWindow {
...
+  org: string;
+  testTimestamp?: string;
+}
+
+export function calculateTimeToRespondBy({
+  numDays,
+  productArea,
+  repo,
+  org,
+  testTimestamp,
```
azaslavsky, line 41: nit: Is this an argument that we only pass in for testing? I wonder if there's another way to do this, a la pythong's freezgun.

azaslavsky (approved):

PR #747, opened 2024-01-25
Resolve some consecutive http queries

hubertdeng123:


azaslavsky (approved):

IanWoodard (approved):
Nice 👍🏻. One thing to watch out for is that if any of the promises reject, Promise.all rejects them all. With that said, I think we handle failures very gracefully so this shouldn't be a problem.

hubertdeng123:
> Nice 👍🏻. One thing to watch out for is that if any of the promises reject, Promise.all rejects them all. With that said, I think we handle failures very gracefully so this shouldn't be a problem.

Looks like I can use `Promise.allSettled` so it doesn't immediately reject

PR #748, opened 2024-01-25
Fix a n plus one query

hubertdeng123:


azaslavsky (approved):
Are there any call sids that need to be changed? If not, LGTM.

hubertdeng123:
> Are there any call sids that need to be changed?

Nope, that should be the only one

PR #756, opened 2024-02-12
Adding soak time failure alert to slack

IanWoodard:
We are rolling out an update to deployments which will cause deploys to fail during soak time if problematic signals are detected. With this comes the need for an alert for engineers to know that soak time failed, specifically those with changes that are in the failed deploy. This change builds off the existing logic used to alert engineers on canary failures.

### DIFF
--- a/src/brain/gocdSlackFeeds/index.ts
+++ b/src/brain/gocdSlackFeeds/index.ts
@@ -1,3 +1,5 @@
+import * as Sentry from '@sentry/node';
+
 import { getUser } from '@/api/getUser';
 import { getAuthors } from '@/api/github/getAuthors';
 import { gocdevents } from '@/api/gocdevents';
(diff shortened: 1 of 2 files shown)


src/brain/gocdSlackFeeds/index.ts
```diff
@@ -128,43 +181,44 @@ const engineeringFeed = new DeployFeed({
...
 
     const blocks = [
-      header(plaintext(':double_vertical_bar: Canary has been paused')),
+      header(
+        plaintext(`:double_vertical_bar: ${pipeline.name} has been paused`)
+      ),
       section(
-        markdown(`The deployment pipeline has been paused due to detected issues in canary. Here are the steps you should follow to address the situation:\n
-:mag_right: *Step 1: Review the Errors*\n Review the errors in the *<${gocdLogsLink}|Canary Logs>*.\n
+        markdown(`The deployment pipeline has been paused due to detected issues in ${pauseCause}. Here are the steps you should follow to address the situation:\n
```
hubertdeng123, line 220: Does it make sense to say `The deployment pipeline has been paused due to detected issues in soak-time`? Is soak-time in s4s? I am missing context on soak time but guessing that other devs would too

IanWoodard: Soak time is the period following a deploy that we let the recently deployed changes sit and "soak" while we collect info about potential problems. It is basically post-deploy monitoring. I use the term "soak-time" as that is what is used in the pipelines and I wanted to keep naming consistent.

IanWoodard: With that said, I could add more/clearer information about soak time in the message

hubertdeng123: Ah, it is a stage in the deployment pipeline entirely. Ok, that makes sense

hubertdeng123 (approved):

PR #765, opened 2024-03-18
Refactored sentry-options to handle more than 10 blocks

kneeyo1:


src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -205,23 +121,84 @@ export async function messageSlack(message: SentryOptionsResponse) {
...
-  for (let i = 0; i < block.length; i += 10) {
-    splitBlock.push(block.slice(i, i + 10));
+
+function createOptionBlocks(
+  options: any[],
+  option_type: string,
+  formatter: OptionFormatter
+): KnownBlock[] {
+  const block: KnownBlock[] = [];
+  const header = `*${option_type.charAt(0).toUpperCase()}${option_type.slice(
```
hubertdeng123, line 185: Won't this also include the `_` in the option types?

src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -205,23 +121,84 @@ export async function messageSlack(message: SentryOptionsResponse) {
...
+    1
+  )} Options:* `;
+  block.push(slackblocks.section(slackblocks.markdown(header)));
+  const batched_options: MrkdwnElement[] = [];
+  for (let count = 0; count < options.length; count += MAX_BLOCK_SIZE) {
+    for (
+      let curr_batch = 0;
+      curr_batch < Math.min(options.length - count, MAX_BLOCK_SIZE);
+      curr_batch += 1
```
hubertdeng123, line 194: nit: semicolon

src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -205,23 +121,84 @@ export async function messageSlack(message: SentryOptionsResponse) {
...
+   * This function also builds blocks around the options in batches of MAX_BLOCK_SIZE. Each
+   * sectionBlock has a block limit.
+   *
+   */
+
+  if (options.length === 0) {
+    return [];
+  }
+
+  const formatterMap: { [key: string]: OptionFormatter } = {
```
hubertdeng123, line 150: nit: Should this live upstream somewhere? Given that `generateBlock` is called multiple times, meaning this formatterMap will be recreated on each call

src/webhooks/sentry-options/sentry-options.ts
kneeyo1: yeah, it will. I think that's fine though, it's still very readable, parsing it out for some of the cases doesn't seem necessary to me.

src/webhooks/sentry-options/sentry-options.ts
kneeyo1: I moved this (and the relevant functions) to be better scoped.

hubertdeng123 (approved):

PR #769, opened 2024-03-26
Adding GoCD feed foundation for ingest

IanWoodard:
This PR adds the initial functionality that will enable a message to be sent in #discuss-ingest each time either relay (soon to be renamed, we will have to circle back and rename it here once that is done) and relay-pop are deployed. From here, you all are free to extend the functionality to fit your needs.

### DIFF
--- a/src/brain/gocdSlackFeeds/index.ts
+++ b/src/brain/gocdSlackFeeds/index.ts
@@ -14,6 +14,7 @@ import {
   FEED_DEPLOY_CHANNEL_ID,
   FEED_DEV_INFRA_CHANNEL_ID,
   FEED_ENGINEERING_CHANNEL_ID,
+  FEED_INGEST_CHANNEL_ID,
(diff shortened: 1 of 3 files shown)


iker-barriocanal (commented):
What does the message look like? Is there a mechanism to update the message contents?

jjbayer (approved):
Thanks for this! Let's try it.

Dav1dde (approved):
Let's try it, but would also be interested in maybe customizing the message.

olksdr (approved):

olksdr:
> Let's try it, but would also be interested in maybe customizing the message.

@Dav1dde  you definitely wanna have more emoji, don't you?

Dav1dde:
> @Dav1dde you definitely wanna have more emoji, don't you?

We need some randomness, random selection of emojis with random counts. Kind of like an oracle, so we can know how the deployment will go before it's even rolled out.

PR #770, opened 2024-03-26
Add header to each section of batched options

kneeyo1:


src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -130,9 +130,9 @@ export async function messageSlack(message: SentryOptionsResponse) {
     const header = `*${option_type.charAt(0).toUpperCase()}${option_type.slice(
       1
     )} Options:* `;
-    block.push(slackblocks.section(slackblocks.markdown(header)));
-    const batched_options: MrkdwnElement[] = [];
     for (let count = 0; count < options.length; count += MAX_BLOCK_SIZE) {
+      block.push(slackblocks.section(slackblocks.markdown(header)));
+      const batched_options: MrkdwnElement[] = [];
```
hubertdeng123, line 135: Do we need to be concerned around `batched_options` being moved into the for loop and being set to [] on every iteration? That seems like it might lead to different behavior from what it was previously so just wanted to make sure

kneeyo1: Thats fine, I think it actually belongs in here and should be reinitalized everytime so it's empty, might have been a bug.

hubertdeng123 (approved):

PR #771, opened 2024-03-26
Fix/sentry options title block

kneeyo1:


src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -134,8 +134,8 @@ export async function messageSlack(message: SentryOptionsResponse) {
       block.push(slackblocks.section(slackblocks.markdown(header)));
       const batched_options: MrkdwnElement[] = [];
       for (
-        let curr_batch = 0;
-        curr_batch < Math.min(options.length - count, MAX_BLOCK_SIZE);
+        let curr_batch = count;
```
hubertdeng123, line 137: Would it be helpful to write a test here to prevent future regressions?

hubertdeng123 (approved):

Issue #773, opened 2024-03-26
Consolidate getsentry pipeline constants for various regions


IanWoodard:
Currently we only have a single constant for getsentry backend pipeline names, which is disingenuous as US is not the only region as we also have s4s and de (and other single tenants). Ideally, we should better utilize pipedream to create a better experience for navigating gocd pipeline names using newer typescript features.

PR #774, opened 2024-03-26
Adding de to the team-engineering gocd feed

IanWoodard:
Now that DE comes after S4S and before US, we should make sure it alerts when it fails to prevent silent failures that go unnoticed. I also created an issue to update/refactor the constants for GoCD pipeline names, but given that I want this out before the end of the day, I opted to punt this for later.
https://github.com/getsentry/eng-pipes/issues/773


kneeyo1 (approved):
LGTM

PR #785, opened 2024-05-03
ref: update gocd failure message + sns feeds

kylemumma:
This PR:
1. Changes the GoCD deploy failure message to 'GoCD auto-deployment failed'
2. make feed-sns only get pipeline fail message and not start message

### DIFF
--- a/src/brain/gocdSlackFeeds/deployFeed.ts
+++ b/src/brain/gocdSlackFeeds/deployFeed.ts
@@ -298,7 +298,12 @@ export class DeployFeed {
   }
 
   async getBodyText(pipeline: GoCDPipeline) {
-    let body = `GoCD deployment started`;
+    let body  = '';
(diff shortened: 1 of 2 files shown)


src/brain/gocdSlackFeeds/deployFeed.ts
```diff
@@ -298,7 +298,12 @@ export class DeployFeed {
   }
 
   async getBodyText(pipeline: GoCDPipeline) {
-    let body = `GoCD deployment started`;
+    let body  = '';
+    if (pipeline.stage.result.toLowerCase() == 'failed') {
```
kylemumma, line 302: It seems that when one of these deploy messages is posted to slack it starts off as unknown then eventually gets to passed or failed. So I'm concerned that when this function is triggered it will have result unknown, later change to failed, but when it changes to failed not trigger the function again, thus the fail message never gets posted....

PR #790, opened 2024-05-10
feat(sns): Add a deploy feed in discuss-eng-sns when pipeline fails

john-z-yang:
Add a deploy feed in discuss-eng-sns when pipeline fails.
When any stage of the deploy-snuba-us or deploy-snuba-de deployment pipeline fails, a message will be sent out in #discuss-eng-sns pining the commit authors.

### DIFF
--- a/src/brain/gocdSlackFeeds/index.ts
+++ b/src/brain/gocdSlackFeeds/index.ts
@@ -211,10 +211,10 @@ const discussSnSFeed = new DeployFeed({
     const gocdLogsLink = `https://deploy.getsentry.net/go/tab/build/detail/${pipeline.name}/${pipeline.counter}/${pipeline.stage.name}/${pipeline.stage.counter}/${failedJob.name}`;
 
     const blocks = [
-      header(plaintext(`:double_vertical_bar: ${pipeline.name} has failed`)),
+      header(plaintext(`:x: ${pipeline.name} has failed`)),
       section(
         markdown(`The deployment pipeline has failed due to detected issues in ${pipeline.stage.name}.\n
-Review the Errors*\n Review the errors in the *<${gocdLogsLink}|GoCD Logs>*.`)
+*Review the errors* in the *<${gocdLogsLink}|GoCD Logs>*.`)
       ),
     ];
     if (ccUsers.length > 0) {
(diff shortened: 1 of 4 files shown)


src/brain/gocdSlackFeeds/index.ts
```diff
@@ -169,6 +170,68 @@ const snsSaaSFeed = new DeployFeed({
...
+      // This should never happen, but if it does, we want to know about it
+      Sentry.captureException(
+        new Error('Failed to find failed job in failed pipeline')
+      );
+      return [];
+    }
+    const gocdLogsLink = `https://deploy.getsentry.net/go/tab/build/detail/${pipeline.name}/${pipeline.counter}/${pipeline.stage.name}/${pipeline.stage.counter}/${failedJob.name}`;
+
+    const blocks = [
+      header(plaintext(`:double_vertical_bar: ${pipeline.name} has failed`)),
```
IanWoodard, line 214: You may not want to use the double_vertical_bar here as that represents paused. I'd recommend using something like `:x:` instead.

src/brain/gocdSlackFeeds/index.ts
```diff
@@ -169,6 +170,68 @@ const snsSaaSFeed = new DeployFeed({
...
+      section(
+        markdown(`The deployment pipeline has failed due to detected issues in ${pipeline.stage.name}.\n
+Review the Errors*\n Review the errors in the *<${gocdLogsLink}|GoCD Logs>*.`)
```
IanWoodard, line 217:
```suggestion
*Review the Errors*\n Review the errors in the *<${gocdLogsLink}|GoCD Logs>*.`)
```
I think you want the extra asterisk here.

IanWoodard: You may even want to remove the label as it looks kind of repetitive with due to there only being one item in the list rather than the 5 items in the list for getsentry.

IanWoodard (approved):
Looks good overall. There are definetly some places where we could further abstract the code, but I don't want to block you on that. I'll make in GitHub issue in eng-pipes about it so we can come back to it later.

src/brain/gocdSlackFeeds/index.ts
kylemumma: done

volokluev (approved):

PR #791, opened 2024-05-13
Fix wild issue slack notification

hubertdeng123:
This was using the previous system of storing product area label to channel mappings in postgres, which is now dated. This brings this up to date with the product-owners.yml file that maps all product areas to teams, which then maps to channels.

### DIFF
--- a/src/utils/getChannelsForIssue.ts
+++ b/src/utils/getChannelsForIssue.ts
@@ -1,3 +1,4 @@
+import * as Sentry from '@sentry/node';
 import moment from 'moment-timezone';
 
 import { PRODUCT_OWNERS_INFO } from '@/config';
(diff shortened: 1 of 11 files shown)


src/utils/getChannelsForIssue.ts
```diff
@@ -0,0 +1,36 @@
...
+): ChannelItem[] => {
+  const teams = getTeams(repo, org, productArea);
+  if (!teams.length) {
+    return [];
+  }
+  const channels: ChannelItem[] = teams.reduce(
+    (acc: ChannelItem[], team: string) => {
+      const offices = PRODUCT_OWNERS_INFO['teams'][team]['offices'];
+      acc.push({
+        channelId: PRODUCT_OWNERS_INFO['teams'][team]['slack_channel'],
```
IanWoodard, line 26: Should we handle the case where PRODUCT_OWNERS_INFO['teams'][team]['offices'] and or PRODUCT_OWNERS_INFO['teams'][team]['slack_channel'] doesn't exist?

IanWoodard (approved):
One comment but otherwise, lgtm

PR #798, opened 2024-06-14
feat: Add gocd paused pipeline bot

IanWoodard:
This PR adds a bot that runs periodically and checks the state pipelines in GoCD and sends slack notifications if specific ones are left paused for extended periods of time. To accomplish this, I needed to introduce a means of authenticating via OIDC in order to work with IAP.


package.json
```diff
@@ -62,11 +62,13 @@
     "@types/tar": "^6.1.4",
     "fastify": "^3.29.4",
     "fastify-formbody": "^5.2.0",
+    "google-auth-library": "^9.10.0",
```
hubertdeng123, line 65: haha I added to into my PR as well

package.json
```diff
@@ -62,11 +62,13 @@
...
     "js-yaml": "^4.1.0",
     "knex": "^2.4.2",
     "middie": "^5.3.0",
     "module-alias": "^2.2.2",
     "moment-timezone": "^0.5.39",
+    "node-fetch": "2",
```
hubertdeng123, line 71: is it possible to use a later version here?

src/utils/iap.ts
```diff
@@ -0,0 +1,11 @@
+import { GoogleAuth } from 'google-auth-library';
+
+import { IAP_TARGET_AUDIENCE } from '@/config';
+
+const auth = new GoogleAuth();
+
+export async function getIDToken(): Promise<string> {
+  const client = await auth.getIdTokenClient(IAP_TARGET_AUDIENCE);
```
hubertdeng123, line 8: I think you can use the client here to make requests instead of using fetch actually, so we can use one less library 🤔

src/webhooks/pubsub/gocdPausedPipelineBot.ts
```diff
@@ -0,0 +1,135 @@
...
+  try {
+    dashboardResult = await fetchDashboard();
+  } catch (err) {
+    Sentry.captureException(err);
+    return;
+  }
+  const remindersByChannel = getRemindersByChannel(dashboardResult, now);
+  for (const [channel, pausedPipelineInfos] of remindersByChannel.entries()) {
+    const pausedPipelineReminderText = getReminderText(pausedPipelineInfos);
+    postMessageToSlack(
```
hubertdeng123, line 41: Should we use await/async here? To ensure behavior is consistent with what we want and nothing appears out of order for any reason

hubertdeng123 (commented):
Well written code! Was super easy to read and understand

src/utils/iap.ts
IanWoodard: The problem I ran into is that when I used the client, it would inject the id token as an authorization instead of proxy-authorization. I can look into this some more though

src/webhooks/pubsub/gocdPausedPipelineBot.ts
IanWoodard: Good catch

package.json
IanWoodard: I had issues with newer versions as our code isn't fully compatible with ESM-only modules and v2 is compatible with CommonJS.

IanWoodard: Bumped it to the latest version that is supported

hubertdeng123 (approved):

PR #799, opened 2024-06-14
Auto route issues

hubertdeng123:
Uses a cloud run instance running a server that is able to make predictions from issue text. You can find out more about the cloud run instance in the repo variables `PREDICT_ENDPOINT`. I've tested this locally by hitting the endpoint on the cloud run instance.

Few things to note:
1. We will only auto route issues that have a probability of predicted class is >= 0.7. 
2. Google Auth is required in order to authenticate predict requests to the server
3. We'll store all these auto routed issues in a db so that in a few months, we can write a script to check how many of the issues are actually routed correctly. This can be done by comparing the predicted product area and the final product area on an issue.

GCP setup for the server is not productionalized, but wanted to get something out to see if it's working before working on that a bit more. 

Screenshot 2024-06-14 at 11 04 28 AM


src/brain/issueLabelHandler/index.test.ts
```diff
@@ -797,7 +877,7 @@ describe('issueLabelHandler', function () {
       }
     );
 
-    it.only.each([
```
hubertdeng123, line 800: 😬

src/brain/issueLabelHandler/route.ts
```diff
@@ -40,14 +51,65 @@ function shouldLabelBeRemoved(labelName, target_name) {
...
+      method: 'POST',
+      data: { text: issueText },
+    });
+    // @ts-ignore Response is of type unknown since it comes from inference service
+    const inferenceJSON: any = inferenceResponse?.data;
+    predictedLabel = inferenceJSON.predicted_label;
+    const probability = inferenceJSON.probability;
+    // Only auto-route if probability is 0.7 or over
+    if (probability >= 0.7) {
+      await org.api.issues.addLabels({
```
hubertdeng123, line 77: These lines are definitely covered by tests...lol

src/brain/issueLabelHandler/index.test.ts
azaslavsky: ?

hubertdeng123: We were skipping tests, I must've accidentally committed this while running tests locally

azaslavsky:
Is there any way to enable this on a subset of repos, or with a higher confidence to start? I worry that it will end up producing too much noise, "training" folks to ignore it in the manner we discussed yesterday. What if we picked a small number of canary repos, and set confidence to like 0.85 or 0.9? It probably wouldn't make too many auto-routings, but that will give us a small cohort to get feedback on.

hubertdeng123:
> Is there any way to enable this on a subset of repos, or with a higher confidence to start? I worry that it will end up producing too much noise, "training" folks to ignore it in the manner we discussed yesterday. 

This will only auto route on sentry and sentry-docs. So, I don't think we can really get a subset of repos smaller than this. 

> What if we picked a small number of canary repos, and set confidence to like 0.85 or 0.9? It probably wouldn't make too many auto-routings, but that will give us a small cohort to get feedback on.

I'm open to raising the confidence and getting the feedback on the auto-routings from those issues. I'll raise it to 0.9, and we can then lower the threshold based on the results we see.

azaslavsky (approved):

Issue #633, opened 2023-09-27
Remove usage of service keys


hubertdeng123:
Instead of service keys, let's use workload identity for better security for our bot.
```[tasklist]
### Tasks
- [x] Create a GCP service account in the security-as-code repo to be used by OIDC (https://github.com/getsentry/security-as-code/pull/384)
- [x] Grant the service account access to deploy Cloud Run services on the target GCP project (`roles/run.developer` or `role/run.admin` which ever is least privilege for the requirements)
- [x] Update `.github/workflows/ci.yml` to use Google's "keyless" auth GHA (https://github.com/getsentry/eng-pipes/pull/647)
- [x] Create and assign a dedicated service account for the Cloud Run service to run-as (https://github.com/getsentry/ops/pull/9016)
- [x] Grant the service account access to read/write/run jobs from google BigQuery (https://github.com/getsentry/ops/pull/9121)
```

PR #800, opened 2024-06-18
fix: Migrations script

hubertdeng123:
Looks like this service account no longer exists. We've been using workload identity provider since https://github.com/getsentry/eng-pipes/issues/633 but we forgot to account for migrations.

https://github.com/getsentry/eng-pipes/actions/runs/9573265261/job/26394333363

### DIFF
--- a/.github/workflows/migration.yml
+++ b/.github/workflows/migration.yml
@@ -22,6 +22,9 @@ jobs:
     name: run migration
     runs-on: ubuntu-latest
     environment: migration
+    permissions:
+      id-token: write
+      contents: read
 
     steps:
     - name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/migration.yml
```diff
@@ -34,13 +34,15 @@ jobs:
...
         node-version: '12'
 
-    # Setup gcloud CLI
-    - name: Setup gcloud CLI
-      uses: google-github-actions/setup-gcloud@v0
-      with:
-        service_account_email: ${{ secrets.GOOGLE_SERVICE_ACCOUNT_EMAIL }}
-        service_account_key: ${{ secrets.GOOGLE_APPLICATION_CREDENTIALS }}
-        export_default_credentials: true
+    - id: 'auth'
```
hubertdeng123, line 37: switching over to using identity provider as done in ci.yml

https://github.com/getsentry/eng-pipes/blob/04f8a2adeb0de29f3a7f0d0675cd4be91dbef55e/.github/workflows/ci.yml#L125

.github/workflows/migration.yml
```diff
@@ -34,13 +37,15 @@ jobs:
       with:
         node-version: '12'
```
IanWoodard, line 38: We should probably update this to use node 20, which may require updating the rest of the actions in this file, but will help us ensure we are on the latest, most secure, version. This can be done in a separate PR.

IanWoodard (approved):
We should opt to use node 20 now that previous versions are deprecated on GitHub actions, but that can be done in a separate PR. Otherwise, LGTM.

PR #810, opened 2024-07-01
fix: Make string truncation unicode aware

BYK:


.github/workflows/ci.yml
```diff
@@ -112,9 +112,9 @@ jobs:
           fetch-depth: 0
 
       - name: Setup node
-        uses: actions/setup-node@v3
+        uses: actions/setup-node@v4
         with:
-          node-version: '14'
```
hubertdeng123, line 117: I think we can bump these versions in `migration.yml` as well

hubertdeng123 (approved):

.github/workflows/ci.yml
BYK: #811

PR #818, opened 2024-07-26
feat: add slack notifications from infra-event-notifier

bmckerry:
This adds webhooks for the [infra-event-notifier](https://github.com/getsentry/infra-event-notifier) service to eng-pipes.

It implements HMAC signing for incoming requests, which on the client side will have a signing secret loaded into the infra-event-notifier via an environment variable.


src/webhooks/kafka-control-plane/kafka-control-plane.ts
```diff
@@ -0,0 +1,108 @@
...
+      payloadBody,
+      clientSignature!,
+      KAFKA_CONTROL_PLANE_WEBHOOK_SECRET!,
+      (i) => i,
+      'sha256'
+    );
+
+    if (!isVerified) {
+      return reply.code(401).send('Unauthorized');
+    }
```
IanWoodard, line 37: I'd recommend utilizing [extractAndVerifySignature](https://github.com/getsentry/eng-pipes/blob/fb8c72ff8d4939f5736388810b4ba9124d397195/src/utils/extractAndVerifySignature.ts#L7) to reduce code duplication similar to how it is used in the gocd webhook https://github.com/getsentry/eng-pipes/blob/fb8c72ff8d4939f5736388810b4ba9124d397195/src/webhooks/gocd/gocd.ts#L19-L29.

src/config/index.ts
```diff
@@ -228,6 +230,12 @@ export const GOCD_PAUSED_PIPELINE_REMINDERS: GoCDPausedPipelineReminder[] = [
 export const SENTRY_OPTIONS_WEBHOOK_SECRET =
   process.env.SENTRY_OPTIONS_WEBHOOK_SECRET;
 export const GOCD_WEBHOOK_SECRET = process.env.GOCD_WEBHOOK_SECRET;
+export const KAFKA_CONTROL_PLANE_WEBHOOK_SECRET =
+  process.env.KAFKA_CONTROL_PLANE_WEBHOOK_SECRET;
+
+if (KAFKA_CONTROL_PLANE_WEBHOOK_SECRET === undefined) {
```
IanWoodard, line 236: Instead of bricking eng-pipes if this secret isn't set, could we just send respond with 500 on the specific endpoint?

PR #823, opened 2024-07-30
fix(gocd-webhook): adding hmac auth to GoCD webhook

IanWoodard:
Adds hmac auth to the gocd webhook. This uses the newly added GOCD_WEBHOOK_SECRET to generate a signature and compare it to the signature provided by the webhook requests.


src/config/index.ts
```diff
@@ -219,3 +219,8 @@ export const GOCD_PAUSED_PIPELINE_REMINDERS: GoCDPausedPipelineReminder[] = [
     notifyAfter: moment.duration(1.5, 'hour'),
   },
 ];
+
+/**
+ * GoCD Webhook Secret
+ */
+export const GOCD_WEBHOOK_SECRET = process.env.GOCD_WEBHOOK_SECRET || '';
```
mdtro, line 226: We shouldn't default to an empty string here. Startup should fail if the value is not set, otherwise we will silently be hashing things with an empty string (which is very predictable).

src/utils/verifySignature.ts
```diff
@@ -6,9 +6,10 @@ export function verifySignature(
   payload: string,
   signature: string,
   secret: string,
-  createDigest = (i) => i
+  createDigest = (i) => i,
+  method: 'sha1' | 'sha256' = 'sha1'
```
mdtro, line 10: Can we default to the stronger algorithm, `sha256`?

src/webhooks/gocd/gocd.ts
```diff
@@ -1,12 +1,40 @@
...
-  return {};
-  const { body }: { body: GoCDResponse } = request;
-  gocdevents.emit(body.type, body);
-  return {};
+export async function handler(
+  request: FastifyRequest<{ Body: GoCDResponse }>,
+  reply: FastifyReply
+) {
+  try {
+    const clientSignatureHeader = request.headers['x-gocd-signature'] ?? '';
```
mdtro, line 15: Should immediately return a 400 or 401 if this header is empty.

src/utils/verifySignature.ts
IanWoodard: The function this function calls defaults to sha1, I can change all existing usages that rely on that default to be explicit and change the default to sha256.

mdtro (approved):

src/webhooks/gocd/gocd.ts
```diff
@@ -1,12 +1,37 @@
...
-  const { body }: { body: GoCDResponse } = request;
-  gocdevents.emit(body.type, body);
-  return {};
+export async function handler(
+  request: FastifyRequest<{ Body: GoCDResponse }>,
+  reply: FastifyReply
+) {
+  // If the webhook secret is not defined, return a 500
+  if (GOCD_WEBHOOK_SECRET === undefined) {
+    return reply.code(500).send();
```
hubertdeng123, line 16: should we send a Sentry error if this happens for visbility?

hubertdeng123: Thinking it may be hard to track down if we just return a 500 http response

IanWoodard: Moved it into the try catch and threw an error that will be caught by sentry

PR #825, opened 2024-08-01
Add a webhook for handling slack notifications from status-page webhooks

brian-lou:
This PR adds an endpoint that can send Slack messages to slack channels, to be used as a middle-man for the Sentry internal Status page service


src/slack/status-page/status-page.test.ts
```diff
@@ -0,0 +1,114 @@
...
+        channel: 'AAAAA',
+        unfurl_links: false,
+      });
+    });
+
+    // it('only writes kafka-control-plane changes', async function () {
+    //   const postMessageSpy = jest.spyOn(bolt.client.chat, 'postMessage');
+    //   await messageSlack(testAdminPayload);
+    //   expect(postMessageSpy).toHaveBeenCalledTimes(0);
+    // });
```
IanWoodard, line 112:
```suggestion
```

Remove?

IanWoodard (approved):
LGTM. Lot's of duplicate code from https://github.com/getsentry/eng-pipes/commit/211961ab6ea1a48961764affdf357c91012c06fd#diff-d8a76bbdc99e7610cca3c418e2f7e34b4785fe2b1c5c21c112ce2b2438663875. Maybe we should abstract this logic out to reduce code duplication. This can be done later though, would be good to make an issue so we remember.

PR #828, opened 2024-08-07
Adding a link to the paused pipeline for canary/soak time messages

### LINKED ISSUE #815: Add link to pipeline that is paused in the canary/soak time failure message
At the end of the canary/soak time failure message we mention unpausing the pipeline. We should include a link in this section that takes the user to a page where they can quickly click unpause.

IanWoodard:
Fixes #815.
Adds a link to the pipeline that was paused to enable engineers to more easily navigate and unpause the pipeline once it is safe to do so using the instructions in the slack message.


brian-lou (approved):
LGTM

PR #831, opened 2024-08-08
fix: use right header, remove channel from type

bmckerry:
Forgot to commit this in #818, this removes `channel` from the type (since we're hardcoding it) and also uses the right signature header key.


src/webhooks/kafka-control-plane/kafka-control-plane.test.ts
```diff
@@ -33,7 +33,7 @@ describe('kafka-control-plane webhook', function () {
       method: 'POST',
       url: '/metrics/kafka-control-plane/webhook',
       headers: {
-        'x-kafka-control-plane-signature': 'invalid',
+        'x-infra-event-notifier-signature': 'invalid',
```
brian-lou, line 36: I think in infra-event-notifier I have it using the old kafka header. Since the endpoint is called kafka-control-plane, should we keep the old header name?

src/webhooks/kafka-control-plane/kafka-control-plane.ts
```diff
@@ -42,7 +42,7 @@ export async function handler(
 }
 
 export async function messageSlack(message: KafkaControlPlaneResponse) {
-  if (message.source !== 'kafka-control-plane') {
+  if (message.source !== 'infra-event-notifier') {
```
brian-lou, line 45: Same as above

brian-lou (approved):

PR #833, opened 2024-08-14
[OPS-6812] Add run-custom-job replyCallback to produce stage update Slack notifications for GoCD Job Runner

shashjar:


src/brain/gocdSlackFeeds/index.ts
```diff
@@ -366,10 +368,36 @@ const goCDCustomJobRunnerFeed = new DeployFeed({
...
+    }
+
+    const blocks = [
+      header(
+        plaintext(`${pipeline.name} stage update`)
+      ),
+      stageBlock(pipeline),
+    ];
+
+    const approvedBy = pipeline.stage['approved-by'];
```
shashjar, line 386: @IanWoodard does this `approved-by` actually contain the user who kicked off the deployment on the stages in the pipeline after the first one? In the GoCD UI I see "Automatically triggered" instead of "Triggered by <user>".

Not the worst case if we don't have the user to @ in the reply, since they're already tagged in the channel message, so replying in thread at all should re-notify (unless they have this setting turned off?). Safer to @ so would prefer that if possible.

maxwelllwang (approved):

src/brain/gocdSlackFeeds/index.ts
IanWoodard: I'd add some tests for the edge cases here, but it should work for most cases (not the case where it is auto-deployed)

IanWoodard (approved):
Looks good overall, I would add some tests before merging though

src/brain/gocdSlackFeeds/index.test.ts
```diff
@@ -1745,6 +1746,199 @@ Please do not ignore this message just because the environment is not SaaS, beca
...
+                  '<https://deploy.getsentry.net/go/pipelines/run-custom-job/20/checks/1|Failed>'
+                ),
+              ],
+            },
+          ],
+        },
+      ],
+    });
+  });
+
```
IanWoodard, line 1941: What are your thoughts on adding one more test for when the result is not unknown but there is no user?

shashjar: Sure, that's a good idea. Adding now.

IanWoodard (approved):

PR #840, opened 2024-09-05
Updating canary failure message to be more explicit

### LINKED ISSUE #839: Make action for cc'd users in canary failure more explicit
Received some feedback asking if we could make the message to cc'd users more explicit instructing them to triage the situation in the case that canary failed.

IanWoodard:
Fixes #839


mrduncan (approved):
🙌

PR #841, opened 2024-09-05
fix(gocd-job-runner): Fixing approved by to check if valid sentry email

IanWoodard:
Added a check for the gocd-job-runner's deployFeed to ensure we only try to cc a user if the approvedBy is a valid Sentry email.


shashjar (approved):
ty 🙏

PR #844, opened 2024-09-12
Move Slack bot init from api/ into init/

brian-lou:


fpacifici:
Could you please provide some context on the goal?

brian-lou:
Basically atm the goal is to clean up the file structure so that it's cleaner & clearer. My current idea is: https://www.notion.so/sentry/Eng-Pipes-Documentation-74c3b8f9a9ea4b67b3482673b76c7ba7, but we @IanWoodard and I will first fix the tests before we move more things around

PR #849, opened 2024-09-16
Fix Some Tests

brian-lou:
This PR fixes most tests in `utils/` and `webhooks/`

### DIFF
--- a/src/brain/githubMetrics/index.test.ts
+++ b/src/brain/githubMetrics/index.test.ts
@@ -43,7 +43,6 @@ const SCHEMA = Object.entries(dbFunctions.TARGETS.oss.schema).map(
 
 describe('github webhook', function () {
   let fastify: Fastify;
-  // const org = GETSENTRY_ORG;
 
   beforeEach(async function () {
     fastify = await buildServer(false);
@@ -54,7 +53,6 @@ describe('github webhook', function () {
     fastify.close();
     (dbFunctions.insertOss as jest.Mock).mockClear();
     (dbFunctions.insert as jest.Mock).mockClear();
-    // org.api.orgs.checkMembershipForUser.mockClear();
     mockDataset.mockClear();
     mockTable.mockClear();
     mockInsert.mockClear();
--- a/src/brain/githubMetrics/index.test.ts
+++ b/src/brain/githubMetrics/index.test.ts
@@ -22,7 +22,7 @@ import { Fastify } from '@types';
 import { createGitHubEvent } from '@test/utils/github';
 
 import { buildServer } from '@/buildServer';
-import { DRY_RUN, GETSENTRY_ORG } from '@/config';
+import { DRY_RUN } from '@/config';
 import * as dbFunctions from '@utils/metrics';
 
 import { githubMetrics as metrics } from '.';
@@ -43,7 +43,7 @@ const SCHEMA = Object.entries(dbFunctions.TARGETS.oss.schema).map(
 
 describe('github webhook', function () {
   let fastify: Fastify;
-  const org = GETSENTRY_ORG;
+  // const org = GETSENTRY_ORG;
 
   beforeEach(async function () {
     fastify = await buildServer(false);
@@ -54,7 +54,7 @@ describe('github webhook', function () {
     fastify.close();
     (dbFunctions.insertOss as jest.Mock).mockClear();
     (dbFunctions.insert as jest.Mock).mockClear();
-    org.api.orgs.checkMembershipForUser.mockClear();
+    // org.api.orgs.checkMembershipForUser.mockClear();
     mockDataset.mockClear();
     mockTable.mockClear();
     mockInsert.mockClear();
--- a/src/brain/githubMetrics/index.test.ts
+++ b/src/brain/githubMetrics/index.test.ts
@@ -20,9 +20,10 @@ jest.mock('@google-cloud/bigquery', () => ({
 import { Fastify } from '@types';
 
 import { createGitHubEvent } from '@test/utils/github';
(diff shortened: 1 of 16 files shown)


src/brain/githubMetrics/index.test.ts
```diff
@@ -43,7 +43,7 @@ const SCHEMA = Object.entries(dbFunctions.TARGETS.oss.schema).map(
 
 describe('github webhook', function () {
   let fastify: Fastify;
-  const org = GETSENTRY_ORG;
+  // const org = GETSENTRY_ORG;
```
IanWoodard, line 46:
```suggestion
```

We should probably just remove it

src/brain/githubMetrics/index.test.ts
```diff
@@ -54,7 +54,7 @@ describe('github webhook', function () {
     fastify.close();
     (dbFunctions.insertOss as jest.Mock).mockClear();
     (dbFunctions.insert as jest.Mock).mockClear();
-    org.api.orgs.checkMembershipForUser.mockClear();
+    // org.api.orgs.checkMembershipForUser.mockClear();
```
IanWoodard, line 57:
```suggestion
```

IanWoodard (commented):
Some initial comments, will leave some more when I finish the review

src/webhooks/pubsub/slackScores.test.ts
```diff
@@ -65,7 +66,7 @@ describe('slackScores tests', function () {
         issues: [],
         gitHubCommenters: [],
       });
-      await triggerSlackScores(GETSENTRY_ORG, null);
+      await triggerSlackScores(GETSENTRY_ORG, null as any);
```
IanWoodard, line 69: Same as above

src/webhooks/pubsub/index.test.ts
```diff
@@ -37,8 +44,8 @@ describe('slack app', function () {
       headers: {
         authorization: 'Bearer 1234abcd',
       },
-    };
-    const reply = new MockReply();
+    } as FastifyRequest<{ Body: { message: { data: string } } }>;
+    const reply = new MockReply() as any;
```
IanWoodard, line 48: What if we cast it as a `FastifyReply` rather than any. Either way it isn't technically right, but at least with `FastifyReply`, we are less broad.

src/webhooks/pubsub/index.test.ts
```diff
@@ -129,8 +136,8 @@ describe('slack app', function () {
       headers: {
         authorization: 'invalid',
       },
-    };
-    const reply = new MockReply();
+    } as FastifyRequest<{ Body: { message: { data: string } } }>;
+    const reply = new MockReply() as any;
```
IanWoodard, line 140: Same as above

src/webhooks/pubsub/index.test.ts
```diff
@@ -111,8 +118,8 @@ describe('slack app', function () {
         },
       },
       headers: {},
-    };
-    const reply = new MockReply();
+    } as FastifyRequest<{ Body: { message: { data: string } } }>;
+    const reply = new MockReply() as any;
```
IanWoodard, line 122: Same as above

src/webhooks/pubsub/index.test.ts
```diff
@@ -37,8 +44,8 @@ describe('slack app', function () {
       headers: {
         authorization: 'Bearer 1234abcd',
       },
-    };
-    const reply = new MockReply();
+    } as FastifyRequest<{ Body: { message: { data: string } } }>;
+    const reply = new MockReply() as any;
     await pubSubHandler(request, reply);
     return reply;
```
IanWoodard, line 50: I think we also need to update the function definition to include a return type of `Promise<MockReply>` (since it is a subset of FastifyReply)

IanWoodard (commented):
Left a few comments, but it is looking good so far!

src/webhooks/pubsub/index.test.ts
brian-lou: I don't think we should be changing the function definition just for testing

src/webhooks/pubsub/index.test.ts
brian-lou: The issue is that elsewhere in the code `expect(reply.statusCode).toBe(204);` won't work with the FastifyReply type

fpacifici:
> This PR fixes most tests in utils/ and webhooks/

Does it mean CI was not passing ?

brian-lou:
> > This PR fixes most tests in utils/ and webhooks/
> 
> Does it mean CI was not passing ?

The tests were passing and correct, but the TS linter/interpreter for the tests was broken

src/brain/requiredChecks/handleNewFailedBuild.ts
```diff
@@ -215,7 +215,7 @@ export async function handleNewFailedBuild({
   // DEBUG: Not sure why this messages sometimes does not get threaded.
   // The Slack API should make this threaded because existingFailureMessage exists
   // eslint-disable-next-line no-console
-  console.log({ existingFailureMessage, followupFailureMessage });
+  // console.log({ existingFailureMessage, followupFailureMessage });
```
IanWoodard, line 218: Should we keep this around?

IanWoodard (approved):

src/brain/requiredChecks/handleNewFailedBuild.ts
brian-lou: Isn't it bad practice to have console logging in prod? (I'm not sure why the log was needed originally)

PR #850, opened 2024-09-16
fix(tests): Cleaning up gocdHelpers and tests

IanWoodard:
Updated gocdHelpers to use more specific types (using Typescript's Pick operator) to make mocking data easier. Fixed some but not all issues with the gocdHelpers tests. The last one is to fix the typing for the "git-configurations" which is required currently, but doesn't exist when the material type is "pipeline" so it shouldn't be in that case. However, this does put us in a better state than before.
(Also updating vscode settings to only run tests "on-demand")

### DIFF
--- a/src/utils/gocdHelpers.ts
+++ b/src/utils/gocdHelpers.ts
@@ -91,12 +91,15 @@ export function getProgressColor(pipeline: GoCDPipeline) {
 }
 
 export function firstGitMaterialSHA(
-  deploy: DBGoCDDeployment | undefined
+  deploy?: Pick<DBGoCDDeployment, 'pipeline_build_cause'>
 ): string | null {
-  if (!deploy) {
+  if (deploy === undefined) {
     return null;
   }
-  if (deploy.pipeline_build_cause.length === 0) {
+  if (
+    deploy.pipeline_build_cause === undefined ||
+    deploy.pipeline_build_cause.length === 0
+  ) {
     return null;
   }
   for (const bc of deploy.pipeline_build_cause) {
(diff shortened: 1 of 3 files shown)


src/utils/gocdHelpers.ts
```diff
@@ -91,12 +91,15 @@ export function getProgressColor(pipeline: GoCDPipeline) {
 }
 
 export function firstGitMaterialSHA(
-  deploy: DBGoCDDeployment | undefined
+  deploy?: Pick<DBGoCDDeployment, 'pipeline_build_cause'>
```
brian-lou, line 94: Are there any downsides to using Pick here? For example, would this not relax TS's type safety checking?

src/utils/gocdHelpers.ts
```diff
@@ -112,11 +115,11 @@ export function firstGitMaterialSHA(
 }
 
 export function filterBuildCauses(
-  pipeline: GoCDPipeline,
+  pipeline: Pick<GoCDPipeline, 'build-cause'>,
```
brian-lou, line 118: Same question as above

src/utils/gocdHelpers.ts
```diff
@@ -135,7 +138,7 @@ export function filterBuildCauses(
 }
 
 export async function getBaseAndHeadCommit(
-  pipeline: GoCDPipeline
+  pipeline: Pick<GoCDPipeline, 'build-cause' | 'group' | 'name'>
```
brian-lou, line 141: Same question as above

src/utils/gocdHelpers.ts
IanWoodard: Since we only need the `pipeline_build_cause` from the `DBGoCDDeployment`, using `Pick` ensures our type definition remains concise and focused. This approach allows us to work with both supersets types (full `DBGoCDDeployment` types) and more specific types, simplifying testing while maintaining type safety. ([Pick docs](https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys))

brian-lou (approved):

PR #855, opened 2024-09-26
Move utility functions from api/ into utils/

brian-lou:
This PR refactors some utility functions from api/ into utils/. Additionally, it combines two gocd folders for better code readability.


IanWoodard (approved):
LGTM

PR #856, opened 2024-10-02
chore(types): refactoring gocd types

IanWoodard:
Moving GoCD related types into their own file to start to declutter the main types file


brian-lou (approved):
LGTM

PR #861, opened 2024-10-03
Add support for subdirs in brain

brian-lou:


src/utils/loadBrain.ts
```diff
@@ -18,25 +20,49 @@ export async function loadBrain() {
...
+
+    if (file.isDirectory()) {
+      const nestedDirs = await getBrainModules(filePath);
+      nestedDirs.forEach((nestedDir) => directories.add(nestedDir));
+    } else if (
+      !file.name.endsWith('.d.ts') &&
+      !file.name.endsWith('.map') &&
+      !file.name.endsWith('.md')
+    ) {
+      directories.add(path.relative(ROOT, dir)); // Add relative path to directory
```
IanWoodard, line 43: This comment doesn't add much value
```suggestion
      directories.add(path.relative(ROOT, dir));
```

src/utils/loadBrain.ts
```diff
@@ -18,25 +20,49 @@ export async function loadBrain() {
...
-    (f) => !f.endsWith('.d.ts') && !f.endsWith('.map') && !f.endsWith('.md')
-  );
+export async function getBrainModules(dir: string = ROOT): Promise<string[]> {
+  // Read the directory contents
+  const files = await fs.readdir(dir, { withFileTypes: true });
+  const directories: Set<string> = new Set();
+
+  // Loop through each file or directory in the current directory
+  for (const file of files) {
+    const filePath = path.join(dir, file.name);
```
IanWoodard, line 33: Nit: Let's move this definition inside of the if statement since it's only used in there. We could even remove it entirely (up to you)

src/utils/loadBrain.ts
```diff
@@ -18,25 +20,49 @@ export async function loadBrain() {
...
+
+    if (file.isDirectory()) {
+      const nestedDirs = await getBrainModules(filePath);
+      nestedDirs.forEach((nestedDir) => directories.add(nestedDir));
+    } else if (
+      !file.name.endsWith('.d.ts') &&
+      !file.name.endsWith('.map') &&
+      !file.name.endsWith('.md')
```
IanWoodard, line 41: This is failing for me locally, for some reason I have a .tox directory (not tracked by git) inside the brain that is causing issues since it's files don't fall under these three filters. Why don't we just check if the file.name ends with `.ts` rather than checking what it doesn't end with (while keeping the check to make sure it doesn't end with `.d.ts` first)?

IanWoodard (approved):

PR #865, opened 2024-10-08
Modify loadBrain to use imports instead of dynamic path loading

brian-lou:
Per the discussion, this PR switches loadBrain to load handlers based off of imports & exported functions rather than based off of modules in the `brain/` path.


IanWoodard (approved):
LGTM

PR #871, opened 2024-10-09
Move brain modules into subdirs

brian-lou:


src/utils/loadBrain.ts
```diff
@@ -1,23 +1,23 @@
...
-import { syncSlackUsers } from '@/brain/syncSlackUsers';
-import { syncUserProfileChange } from '@/brain/syncUserProfileChange';
-import { triggerPubSub } from '@/brain/triggerPubSub';
-import { typescript } from '@/brain/typescript';
+import { apis } from '@/brain/github/apis';
+import { ghaCancel } from '@/brain/github/ghaCancel';
+import { githubMetrics } from '@/brain/github/githubMetrics';
+import { issueLabelHandler } from '@/brain/github/issueLabelHandler';
+import { issueNotifier } from '@/brain/github/issueNotifier';
+import { pleaseDeployNotifier } from '@/brain/github/pleaseDeployNotifier';
```
IanWoodard, line 6: This is a bit of a weird one since it fits into both github and gocd, though I do think it fits better in the former as you have it. Any thoughts?

IanWoodard (approved):

src/utils/loadBrain.ts
brian-lou: The idea is to group the listeners for the upstream sources together, so I think github is right here

PR #872, opened 2024-10-10
Switch cron jobs from pubsub to HTTP triggers

brian-lou:
This PR deprecates pubsub endpoints and switches to HTTP endpoints for triggering cron jobs. This is the first step of the migration. Once this has been merged & is stable, the cron jobs will be redirected away from pubsub. Once this is stable, we will remove the old pubsub endpoints in another PR.


IanWoodard (changes requested):
Let's leave the old way in until the other change is merged so there is no downtime for these cron jobs

src/jobs/index.ts
```diff
@@ -0,0 +1,119 @@
...
+        },
+      },
+    },
+  },
+};
+
+// Function that creates a sub fastify server for job webhooks
+export async function routeJobs(server: Fastify, _options) {
+  server.post('/stale-triage-notifier', (request, reply) =>
+    handleJobRoute(notifyProductOwnersForUntriagedIssues, request, reply)
```
IanWoodard, line 100: Could we get coverage for these lines?

IanWoodard (approved):

PR #874, opened 2024-10-11
Delete old pubsub handlers

brian-lou:
This PR follows up on #872 by deprecating the old pubsub code.


hubertdeng123 (commented):
Looks good to me. What is the plan of rollout here? How will we ensure that existing functionality will not break once cloud scheduler begins sending requests?

IanWoodard (approved):
Looks good since we already updated cloud scheduler

PR #877, opened 2024-10-16
Add Sentry txn observability for webhooks

brian-lou:
This PR adds Sentry txn tracing to the webhook endpoints

### DIFF
--- a/src/webhooks/index.ts
+++ b/src/webhooks/index.ts
@@ -9,38 +9,47 @@ import { kafkactlWebhook } from './kafka-control-plane/kafka-control-plane';
 import { sentryOptionsWebhook } from './sentry-options/sentry-options';
 import { webpackWebhook } from './webpack/webpack';
 
+type WebhookHandler = (
+  request: FastifyRequest<any>,
+  reply: FastifyReply
+) => Promise<void>;
+
 // Error handling wrapper function
 export async function handleRoute(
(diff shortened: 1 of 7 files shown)


src/webhooks/index.ts
```diff
@@ -9,38 +11,47 @@ import { kafkactlWebhook } from './kafka-control-plane/kafka-control-plane';
...
+) => Promise<void>;
+
 // Error handling wrapper function
 export async function handleRoute(
-  handler,
+  handler: WebhookHandler,
   request: FastifyRequest,
-  reply: FastifyReply
+  reply: FastifyReply,
+  name: string
```
IanWoodard, line 24: Something about this makes me wish there was a better way, but this makes sense and works

src/webhooks/index.ts
```diff
@@ -1,3 +1,5 @@
+import '@sentry/tracing';
```
IanWoodard, line 1: Any idea why this is needed? Is it needed? I know we had this previously, but I am not sure it is needed.

IanWoodard (approved):
Left a few comments, but otherwise lgtm

src/webhooks/index.ts
brian-lou: In order to initialize Sentry this has to be called

PR #879, opened 2024-10-16
Feat: Add a generic webhook for sending event notifications

brian-lou:
This PR adds a generic webhook: `/event-notifier/v1` which can be used to send events and messages to Slack and DD (with support for Jira coming soon).


src/buildServer.ts
```diff
@@ -96,11 +95,5 @@ export async function buildServer(
   // Endpoints for Cloud Scheduler webhooks (Cron Jobs)
   server.register(routeJobs, { prefix: '/jobs' });
 
-  server.post<{ Params: { service: string } }>(
-    '/slack/:service/webhook',
-    {},
-    SlackRouter(server)
-  );
-
```
fpacifici, line 104: Why are you removing this ? Isn't this in use by the existing slack integrations  to receive commands a user can click on a message ?

brian-lou: No, this was something that I previously added for status page notifications, with the intention of using it for future Slack-related things. But now with this PR, that isn't needed anymore

src/webhooks/generic-notifier/generic-notifier.ts
```diff
@@ -0,0 +1,84 @@
...
+    reply.code(500).send();
+    return;
+  }
+}
+
+export async function sendEventToDatadog(
+  message: GenericEvent,
+  timestamp: number
+) {
+  if (message.data.channels.datadog) {
```
fpacifici, line 55: I'd like a pattern where you decide which methods to call in the `genericEventNotifier` function. That is the orchestrating function, which parses the message and trigger the relevant channels. From their name `sendEventToDatadog` and `messageSlack` should always send the message, not decide when not to send.

fpacifici (commented):
Some comments in line

src/types/index.ts
```diff
@@ -26,3 +28,24 @@ export interface KafkaControlPlaneResponse {
...
+    channels: {
+      slack?: string[]; // list of Slack Channels
+      datadog?: string[]; // list of DD Monitors
+      jira?: string[]; // list of Jira Projects
+      bigquery?: string;
+    };
+    tags?: string[]; // Not used for Slack
+    misc: {
+      alertType?: EventAlertType; // Datadog alert type
+      blocks?: (KnownBlock | Block)[]; // Optional Slack blocks
```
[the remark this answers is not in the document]
brian-lou: I think the idea was to have a combination of both: a generic, simple api which a user can use to route a message to a specific channel, and a more complex webhook which has logic to route more complicated events.

I agree that the typing for the generic event payload can be improved, will work on that

src/webhooks/generic-notifier/generic-notifier.ts
```diff
@@ -0,0 +1,84 @@
...
+      'x-infra-hub-signature',
+      EVENT_NOTIFIER_SECRETS[body.source]
+    );
+    if (!isVerified) {
+      // If the signature is not verified, return (since extractAndVerifySignature sends the response)
+      return;
+    }
+
+    await messageSlack(body);
+    await sendEventToDatadog(body, moment().unix());
```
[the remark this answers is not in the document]
brian-lou: The only other one was Jira, but never really got around to getting that added since Alex was (i think?) working a little on the env setup, so I was originally waiting for him to get that done first.

src/types/index.ts
```diff
@@ -26,3 +28,24 @@ export interface KafkaControlPlaneResponse {
   title: string;
   body: string;
 }
+
+export type GenericEvent = {
+  source: string;
+  timestamp: number;
+  service_name?: string; // Official service registry name if applicable
```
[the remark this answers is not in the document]
brian-lou: Yes, the idea was to use the service-registry if the service_name is supplied, but I'm not sure if everything that will use this in the future will be associated with a service. If we can make that guarantee, I can remove the channel parameter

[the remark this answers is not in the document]
fpacifici: > If we can make that guarantee, I can remove the channel parameter
we cannot make this guarantee. 

So the problem is the same as below. Please let's not have an api method with a lot of optional fields at the same level and rely on the user to know which combination are valid and which ones are not. 

Instead, I would suggest one of these two:
- Have more granular methods: one for slack with smart routing that goes to service and one where we specify the channel. Though we need to think through which methods we are going to build upfront.
- Have one method but provide different structured types for the different types of notification: slack with direct routing, slack with service routing (service is relevant only for slack), jira, datadog, etc.

src/types/index.ts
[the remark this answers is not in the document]
fpacifici: Defining the public api of a service is a task that should be thought through carefully as changing it afterwards when clients are using is hard. 
So feel free to take the time to think the details of all the apis method you expect to create and send a PR with only those types and the empty business logic so we can review if the whole api is coherent and meaningful.

src/types/index.ts
```diff
@@ -26,3 +28,39 @@ export interface KafkaControlPlaneResponse {
...
+  type: 'jira';
+  projectId: string;
+  title: string;
+}
+
+export type GenericEvent = {
+  source: string;
+  timestamp: number;
+  data: (DatadogEvent | JiraEvent | SlackMessage | ServiceSlackMessage)[];
+};
```
fpacifici, line 57: Ok, this is better.

src/types/index.ts
```diff
@@ -26,3 +28,39 @@ export interface KafkaControlPlaneResponse {
...
+
+// Currently only used for Slack notifications since
+// service registry only contains Slack channels (and not DD or Jira or others)
+export interface ServiceSlackMessage {
+  type: 'service_notification';
+  service_name: string; // Official service registry service id
+  text: string;
+  blocks?: KnownBlock[] | Block[];
+}
```
fpacifici, line 66: This seems the same as the slack message, except that you can provide the channel via a service name.
Doing it this way would mean we would duplicate all the methods that can take a destination from the service registry.

Why not abstracting the concept of SlackChannel, and pass it here https://github.com/getsentry/eng-pipes/pull/879/files#diff-a9ba9cbedd19c9f66d564d2e89912890209b98f0a7ef19187877d2587300e476R34 .

You can create two implementations: SlackChannel and ServiceSlackChannel. One contains the channel name and the other contains the service name.

src/webhooks/README.md
```diff
@@ -3,6 +3,44 @@
 * Webhooks in "production" are deployed to a Google Cloud Run instance, in the project `super-big-data`. Why? (TODO insert why)
 * The webhook points to `https://product-eng-webhooks-vmrqv3f7nq-uw.a.run.app`
 
+## Generic Event Notifier
+
+The folder `generic-notifier` provides a generic webhook which can be used to send messages to Sentry Slack channels and Sentry Datadog. Using this webhook is VERY simple.
```
fpacifici, line 8: nit, refrain from making judgements about whether the system you built is simple/complex/well designed/etc. 
It does not add much to the content.

PR #881, opened 2024-10-17
feat: Add heartbeat endpoint & tests

brian-lou:
This PR adds an cron job endpoint that, when triggered, sends an cron job heartbeat to Sentry.

### DIFF
--- a/src/jobs/heartbeat.ts
+++ b/src/jobs/heartbeat.ts
@@ -8,13 +8,13 @@ export async function callDatadog(timestamp: number): Promise<void> {
   const params: v1.EventCreateRequest = {
     title: 'Infra Hub Update',
     text: 'Infra Hub is up',
-    alertType: 'error',
+    alertType: 'info',
     dateHappened: timestamp,
     tags: [
-      `source_tool:infra-hub`,
-      `source:infra-hub`,
-      `source_category:infra-tools`,
-      `sentry_user:infra-hub`,
+      'source_tool:infra-hub',
+      'source:infra-hub',
+      'source_category:infra-tools',
+      'sentry_user:infra-hub',
     ],
   };
   await DATADOG_API_INSTANCE.createEvent({ body: params });
--- a/src/jobs/heartbeat.ts
+++ b/src/jobs/heartbeat.ts
@@ -1,14 +1,28 @@
+import { v1 } from '@datadog/datadog-api-client';
 import * as Sentry from '@sentry/node';
+import moment from 'moment-timezone';
 
-import { bolt } from '@/api/slack';
-import { INFRA_HUB_HEARTBEAT_CHANNEL } from '@/config';
+import { DATADOG_API_INSTANCE } from '@/config';
+
+export async function callDatadog(timestamp: number): Promise<void> {
+  const params: v1.EventCreateRequest = {
+    title: 'Infra Hub Update',
+    text: 'Infra Hub is up',
+    alertType: 'error',
+    dateHappened: timestamp,
+    tags: [
+      `source_tool:infra-hub`,
+      `source:infra-hub`,
+      `source_category:infra-tools`,
+      `sentry_user:infra-hub`,
+    ],
+  };
+  await DATADOG_API_INSTANCE.createEvent({ body: params });
+}
 
 export async function heartbeat() {
   try {
-    await bolt.client.chat.postMessage({
-      channel: INFRA_HUB_HEARTBEAT_CHANNEL,
-      text: 'Infra Hub is up',
-    });
+    await callDatadog(moment().unix());
   } catch (err) {
     Sentry.captureException(err);
   }
--- a/src/jobs/heartbeat.ts
+++ b/src/jobs/heartbeat.ts
@@ -1,0 +1,15 @@
+import * as Sentry from '@sentry/node';
+
+import { bolt } from '@/api/slack';
+import { INFRA_HUB_HEARTBEAT_CHANNEL } from '@/config';
+
+export async function heartbeat() {
+  try {
+    await bolt.client.chat.postMessage({
+      channel: INFRA_HUB_HEARTBEAT_CHANNEL,
+      text: 'Infra Hub is up',
+    });
+  } catch (err) {
+    Sentry.captureException(err);
+  }
+}
--- a/src/jobs/heartbeat.ts
+++ b/src/jobs/heartbeat.ts
@@ -21,9 +21,35 @@ export async function callDatadog(timestamp: number): Promise<void> {
 }
 
 export async function heartbeat() {
+  const checkInId = Sentry.captureCheckIn(
+    {
+      monitorSlug: 'infra-hub-heartbeat',
+      status: 'in_progress',
+    },
+    {
+      schedule: {
+        type: 'crontab',
+        value: '*/5 * * * *',
+      },
+      checkinMargin: 1,
+      maxRuntime: 1,
+      timezone: 'America/Los_Angeles',
+    }
+  );
   try {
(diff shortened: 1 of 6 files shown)


src/jobs/index.ts
```diff
@@ -85,25 +86,27 @@ export async function handleGithubJobs(
...
-  },
-};
+// Error handling wrapper function for Cron Jobs
+// Additionally handles Auth from Cloud Scheduler
+export async function handleCronJobs(
+  handler,
+  request: FastifyRequest,
+  reply: FastifyReply
+) {
+  const verified = await handleCronAuth(request, reply);
```
fpacifici, line 96: Does this framework not provide a native way to authenticate requests rather than hoping every developer remembers to add authentication in their application code in every endpoint ?

src/jobs/heartbeat/heartbeat.ts
```diff
@@ -0,0 +1,15 @@
...
+
+import { bolt } from '@/api/slack';
+import { INFRA_HUB_HEARTBEAT_CHANNEL } from '@/config';
+
+export async function heartbeat() {
+  try {
+    await bolt.client.chat.postMessage({
+      channel: INFRA_HUB_HEARTBEAT_CHANNEL,
+      text: 'Infra Hub is up',
+    });
```
fpacifici, line 11: What's the goal of sending this message? 
Is it about testing the Slack integration or are people supposed to notice these messages are not coming through ? I would push back against the second.

src/jobs/index.ts
brian-lou: The auth for jobs are handled via cloud scheduler (managed by terraform), so developers don't need to worry about it

brian-lou: The webhooks do currently require auth for each endpoint, but I'm not sure if there is a way around that from a security perspective, since traffic is open to the internet

fpacifici: What I meant was whether the web framework we use supports that rather than implementing auth inside the application logic.

fpacifici (dismissed):
Please see my comment on the slack message. That is not going to be a robust way to alert people that the system is down. You want a proper datadog alert with pagerduty attached instead.

src/jobs/index.ts
IanWoodard: Fastify, the web framework eng-pipes uses does not (as far as I am aware) natively support hmac auth. With that said, I do think we could potentially do a better job enforcing that webhooks use hmac and reduce the amount of duplication across routes.

src/jobs/index.ts
```diff
@@ -54,7 +52,21 @@ export async function handleJobRoute(
...
     reply.send();
-    return;
+    return false;
+  }
+  return true;
+}
+
+// Error handling wrapper function for Cron Jobs involving Github
+// Additionally handles Auth from Cloud Scheduler
+export async function handleGithubJobs(
```
IanWoodard, line 62: Nit: Could we use singular for the naming here? Also, GitHub is usually written with a capital `H`
```suggestion
export async function handleGitHubJob(
```

src/jobs/index.ts
brian-lou: This PR only touches routes for cron jobs which are all protected so this discussion isn't relevant here, but the main challenge for the normal webhook routes is that they currently each use their own secrets and header names. We shouldn't try to unify the secrets since that's a security vulnerability, but we could try to unify the headers & auth handlers into one function to reduce duplication

src/jobs/index.ts
```diff
@@ -74,39 +86,44 @@ export async function handleJobRoute(
...
-            },
-          },
-        },
-      },
-    },
-  },
-};
+// Error handling wrapper function for Cron Jobs
+// Additionally handles Auth from Cloud Scheduler
+export async function handleCronJobs(
```
IanWoodard, line 91: Nit: Could we use singular for naming here as well? I feel it makes more sense since we are only handling a single cron job here
```suggestion
export async function handleCronJob(
```

src/jobs/index.ts
```diff
@@ -74,39 +86,44 @@ export async function handleJobRoute(
...
-    handleJobRoute(triggerStaleBot, request, reply)
+    handleGithubJobs(triggerStaleBot, request, reply)
   );
   server.post('/slack-scores', (request, reply) =>
-    handleJobRoute(triggerSlackScores, request, reply)
+    handleGithubJobs(triggerSlackScores, request, reply)
   );
   server.post('/gocd-paused-pipeline-bot', (request, reply) =>
-    handleJobRoute(triggerPausedPipelineBot, request, reply)
+    handleGithubJobs(triggerPausedPipelineBot, request, reply)
```
IanWoodard, line 123: While this did use the same handler previously, it would be nice to decouple this job from the GitHub logic since it is unrelated to it and doesn't use the org passed in. The only reason it was written that way originally was for convenience as there was only one mechanism for handling cron jobs.

src/jobs/heartbeat.ts
```diff
@@ -0,0 +1,29 @@
...
+  const params: v1.EventCreateRequest = {
+    title: 'Infra Hub Update',
+    text: 'Infra Hub is up',
+    alertType: 'error',
+    dateHappened: timestamp,
+    tags: [
+      `source_tool:infra-hub`,
+      `source:infra-hub`,
+      `source_category:infra-tools`,
+      `sentry_user:infra-hub`,
```
IanWoodard, line 17: Nit: Is there a reason we are using backticks rather than single quotes here?

src/jobs/heartbeat.ts
```diff
@@ -0,0 +1,29 @@
...
+import * as Sentry from '@sentry/node';
+import moment from 'moment-timezone';
+
+import { DATADOG_API_INSTANCE } from '@/config';
+
+export async function callDatadog(timestamp: number): Promise<void> {
+  const params: v1.EventCreateRequest = {
+    title: 'Infra Hub Update',
+    text: 'Infra Hub is up',
+    alertType: 'error',
```
IanWoodard, line 11: What does an `error` `alertType` imply? I didn't find much info after glancing at their docs.

src/jobs/heartbeat.ts
```diff
@@ -0,0 +1,29 @@
...
+      `source_category:infra-tools`,
+      `sentry_user:infra-hub`,
+    ],
+  };
+  await DATADOG_API_INSTANCE.createEvent({ body: params });
+}
+
+export async function heartbeat() {
+  try {
+    await callDatadog(moment().unix());
```
IanWoodard, line 25: Is the main point of the heartbeat to monitor whether are cron jobs are able to be routed? Shouldn't we try to dogfood and make use of Sentry crons for this since that is basically what we are doing here? Also, this doesn't help us detect if a specific endpoint is broken, just that the routing logic (mostly) works, which is still a good thing to monitor, but it isn't a catch all for all problems.

src/jobs/index.ts
brian-lou: The way the tests are written require both a github org and moment so let's not change it right now

src/jobs/heartbeat.ts
brian-lou: The point of the heartbeat is to monitor if the entire system is reachable / available, or if it's down (like what happened last week).

src/jobs/heartbeat.ts
brian-lou: I think it's just the type of event that is sent

brian-lou:
> Please see my comment on the slack message. That is not going to be a robust way to alert people that the system is down. You want a proper datadog alert with pagerduty attached instead.

Added Sentry cron job monitoring + datadog event alerts

PR #906, opened 2025-04-10
Assign "Product Area: Releases" to Bruno's team

leedongwei:


hubertdeng123:
Hi @leedongwei you will also need to edit this if you'd like this file to stay in sync:

https://github.com/getsentry/security-as-code/blob/787cfc61f2f30559bb742191fa707362c9d8440d/rbac/lib/product-owners.yml#L318

PR #910, opened 2025-05-13
feat(slack-bot): notify original commit authors for failed revert commits

shellmayr:
- Action item from INC-1121
- Add a (default: false) parameter to `getAuthors` to retrieve the authors of original commits in the case of revert
- Add tests
- Use the new function in the Slack feeds for sentry and snuba to notify users when the revert of their commit fails as well

Closes TET-366

### DIFF
--- a/src/utils/github/getAuthors.ts
+++ b/src/utils/github/getAuthors.ts
@@ -74,16 +74,11 @@ async function getRevertCommitDetails(
 
 async function processCommit(
   commitSha: string,
+  commitMessage: string,
   repo: string
 ): Promise<RevertAuthorInfo | null> {
   try {
-    const commitDetailsResponse = await GETSENTRY_ORG.api.repos.getCommit({
-      owner: GETSENTRY_ORG.slug,
-      repo,
-      ref: commitSha,
-    });
-    const commitData = commitDetailsResponse.data;
-    const revertedHash = extractRevertedCommitHash(commitData.commit.message);
+    const revertedHash = extractRevertedCommitHash(commitMessage);
     let revertDetails: RevertAuthorInfo | null = null;
 
     if (revertedHash) {
@@ -118,7 +113,11 @@ export async function getAuthorsWithRevertedCommitAuthors(
 
     const authors = await Promise.all(
       commitsComparison.data.commits.map(async (commitMetadata) => {
-        const revertInfo = await processCommit(commitMetadata.sha, repo);
+        const revertInfo = await processCommit(
+          commitMetadata.sha,
+          commitMetadata.commit.message,
+          repo
+        );
 
         if (revertInfo) {
           return {
--- a/src/utils/github/getAuthors.ts
+++ b/src/utils/github/getAuthors.ts
@@ -58,7 +58,6 @@ export async function getAuthors(
     });
   } catch (err) {
     // eslint-disable-next-line no-console
-    // @ts-ignore
     console.error(err);
     return [];
   }
--- a/src/utils/github/getAuthors.ts
+++ b/src/utils/github/getAuthors.ts
@@ -92,6 +92,28 @@ async function processCommit(
   }
 }
 
+async function processCommitAuthor(
+  commitMetadata: any,
+  repo: string
+): Promise<RevertAuthorInfo> {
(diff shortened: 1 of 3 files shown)


src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+    const revertedData = revertedCommitDetailsResponse.data;
+
+    const revertedCommitAuthor = revertedData.author
+      ? {
+          login: revertedData.author.login,
+          email: revertedData.commit.author?.email ?? null,
+        }
+      : null;
+
+    return revertedCommitAuthor;
```
IanWoodard, line 72: Nit: return above rather than declaring it as a variable.

src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+  try {
+    const revertedCommitDetailsResponse = await octokitInstance.request(
+      'GET /repos/{owner}/{repo}/commits/{ref}',
+      {
+        owner,
+        repo,
+        ref: revertedHash,
+        headers: { 'X-GitHub-Api-Version': '2022-11-28' },
+      }
+    );
```
IanWoodard, line 62: Would it be possible to utilize the higher level octokit apis as we do elsewhere in the code: https://github.com/getsentry/eng-pipes/blob/d4ba7ed6262f9359dc2ae1ce5c01f13db145c48c/src/utils/github/getAuthors.ts#L9 for consistency?

src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+  try {
+    const commitDetailsResponse = await octokitInstance.request(
+      'GET /repos/{owner}/{repo}/commits/{ref}',
+      {
+        owner,
+        repo,
+        ref: commitSha,
+        headers: { 'X-GitHub-Api-Version': '2022-11-28' },
+      }
+    );
```
IanWoodard, line 97: Same question as above

src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+    console.error(`Error processing commit ${commitSha}:`, error);
+    return null;
+  }
+}
+
+export async function getAuthorsWithRevertedCommitAuthors(
+  repo: string,
+  baseCommit: string | null,
+  headCommit: string
+): Promise<Array<{ email: string | undefined; login: string | undefined }>> {
```
IanWoodard, line 122: Why switch to undefined here? Shouldn't we just use the `RevertAuthorInfo` type here and stick to null?

src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+    }
+
+    const authors = await Promise.all(
+      commitsComparison.data.commits.map(async (commitMetadata) => {
+        const revertInfo = await processCommit(
+          commitMetadata.sha,
+          GETSENTRY_ORG.slug,
+          repo,
+          GETSENTRY_ORG.api
+        );
```
IanWoodard, line 144: We should avoid making api calls for each commit as we will quickly hit GitHub's rate limits and run into issues. What if we just extract the messages when from the result of the commit comparison above and use those to determine if we need to fetch info about the revert? Also, we maybe actually be able to get the info about the author of the original commit from that info as well and avoid any extra api all together.

src/utils/github/getAuthors.ts
```diff
@@ -26,6 +28,137 @@ export async function getAuthors(
...
+    }
+
+    return revertDetails;
+  } catch (error) {
+    console.error(`Error processing commit ${commitSha}:`, error);
+    return null;
+  }
+}
+
+export async function getAuthorsWithRevertedCommitAuthors(
```
IanWoodard, line 118: This isn't ever used anywhere. Also, instead of having this be a separate function, we should make it a flag passed into the getAuthors function.

IanWoodard (commented):
I like where this is going, have some nits and suggestions. Let me know if you have any questions!

IanWoodard (approved):

PR #923, opened 2025-08-07
Add options automator latency as a tag and into the msg

kneeyo1:
This adds a latency metric that the options automator can write to datadog to help better report options automator SLOs.

### DIFF
--- a/src/webhooks/sentry-options/sentry-options.test.ts
+++ b/src/webhooks/sentry-options/sentry-options.test.ts
@@ -639,7 +639,7 @@ describe('sentry-options webhook', function () {
     expect(driftedMessage).toEqual({
       body: {
         dateHappened: 1699563828,
-        text: '{"change":"drifted_options","option":{"option_name":"drifted_option_1","option_value":"value_1"}}',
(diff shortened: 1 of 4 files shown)


src/webhooks/sentry-options/sentry-options.ts
```diff
@@ -80,24 +80,34 @@ export async function sendSentryOptionsUpdatesToDatadog(
...
         change: optionType,
         option: option,
+        ...(message.latency_seconds !== undefined && {
+          latency_seconds: message.latency_seconds,
+        }),
       };
 
       const alertType = formatAlertType(optionType);
 
+      const tags = [
```
hubertdeng123, line 90: Could we add a test for this change?

src/webhooks/sentry-options/sentry-options.test.ts
```diff
@@ -628,4 +629,63 @@ describe('sentry-options webhook', function () {
       },
     });
   });
+
+  it('should include latency_seconds tag when present', async function () {
+    await sendSentryOptionsUpdatesToDatadog(testLatencyPayload, 1699563828);
```
hubertdeng123, line 634: nit: might be nice to have this in non-epoch time so it is more readable

PR #924, opened 2025-08-14
Datadog metrics API 

kneeyo1:
This adds the ability to send metrics to datadog. These are different from events and will enable better visibiity on options automator.


hubertdeng123 (approved):
Talked offline, and given time constraints and current design to get this through. A better future would involve sending these metrics from the source

PR #927, opened 2025-09-25
chore(stalebot): Disable stalebot

hubertdeng123:
There's a bug here that I can't repro locally, so disabling until we have some time to look into this

example: https://github.com/getsentry/sentry/pull/86025


hubertdeng123:
If this goes in, I'll create a linear ticket

PR #971, opened 2026-04-14
Trigger a redeploy for secret rotation

geoffg-sentry:
Just a dummy so we can run the deploy script

### DIFF
--- a/bin/deploy.sh
+++ b/bin/deploy.sh
@@ -29,6 +29,7 @@ SENTRY_OPTIONS_WEBHOOK_SECRET
 SENTRY_INFORMER_WEBHOOK_SECRET
 "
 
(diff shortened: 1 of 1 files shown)


bin/deploy.sh
```diff
@@ -29,6 +29,7 @@ SENTRY_OPTIONS_WEBHOOK_SECRET
 SENTRY_INFORMER_WEBHOOK_SECRET
 "
 
+
```
hubertdeng123: this passes linting so good to do, but just for future reference i believe we can actually by using `--allow-empty`

hubertdeng123 (approved):

Jeffreyhung (approved):

